PR #3, opened 2017-11-17
Add OnServerLog hook (for error reporting etc)

kieransouth:
Added to all error types and normal logs too (someone might want to use that in the future)


lukespragg:
Will be merged manually and slightly different in the new repos.

PR #5, opened 2017-11-22
Data now saves on server save/shutdown

Ryan-J-D:


Oxide.Rust/RustCommands.cs
```diff
@@ -663,5 +663,18 @@ private void VersionCommand(IPlayer player, string command, string[] args)
...
+
+        #region Save Command
+
+        [HookMethod("SaveCommand")]
+        private void SaveCommand(IPlayer player, string command, string[] args)
+        {
+            if (!PermissionsLoaded(player)) return;
+
+            Interface.Oxide.OnSave();
+            player.Reply(lang.GetMessage("OxideSaved", this, player.Id));
```
lukespragg: Should not contain "Oxide" in the key, would be better as "SavingData" and "Saving Oxide data..." and likely be good to add a "SavedData" key as well for post save.

Oxide.Rust/Libraries/Covalence/RustPlayerManager.cs
```diff
@@ -35,6 +38,11 @@ internal void Initialize()
             foreach (var pair in playerData) allPlayers.Add(pair.Key, new RustPlayer(pair.Value.Id, pair.Value.Name));
         }
 
+        /// <summary>
+        /// When a player joins
```
lukespragg: I'd probably go with something like... "Register when a player joins" instead, as "When a player joins" doesn't really summarize the method.

Oxide.Rust/Libraries/Covalence/RustPlayerManager.cs
```diff
@@ -53,18 +61,29 @@ internal void PlayerJoin(ulong userId, string name)
                 playerData.Add(id, record);
                 allPlayers.Add(id, new RustPlayer(userId, name));
             }
-
-            ProtoStorage.Save(playerData, "oxide.covalence");
         }
 
+        /// <summary>
+        /// When a player connects
```
lukespragg: I'd probably go with something like... "Register when a player has connected" instead, as "When a player connects" doesn't really summarize the method.

Oxide.Rust/RustCommands.cs
```diff
@@ -663,5 +663,18 @@ private void VersionCommand(IPlayer player, string command, string[] args)
         }
 
         #endregion Version Command
+
+        #region Save Command
+
```
lukespragg: Lacking summary/header info for method. Method should also be placed before Show Command and have a closing #endregion that matches the #region.

Oxide.Rust/RustCore.cs
```diff
@@ -149,14 +150,32 @@ private void OnServerInitialized()
                 Interface.Oxide.LogWarning("The server is currently listed under Community. Please be aware that Facepunch only allows admin tools" +
                                            "(that do not affect gameplay) under the Community section");
 
+            // Custom save timer if server save interval is high (prevents data from never being saved if server crashes)
+            if (ConVar.Server.saveinterval > 3600)
```
lukespragg: The "never" comment here is a bit misleading, as the server could easily crash and lose any data that was lost before 600 seconds. I don't think this is the ideal method to handle this either. It'd be better to just rely on what the server has set for the interval if it has one, otherwise we'd set one for games that don't have this.

lukespragg:
Manually merged without the timer.

PR #18, opened 2017-12-21
Solve a client-side inventory crash when calling Player.ClearInventory()

thibmo:
- Add a new inventory reset method
- Update the clear method to solve a client-side crash


lukespragg:
Merged manually, thanks!

PR #24, opened 2017-12-31
Added 3rd Arg (BasePlayer) to OnItemAction

Jake-Rich:
Without the 3rd argument, there is no way to figure out what player is using an item when the item is in a storage container or another player's inventory.


lukespragg:
Merged manually in latest update, thanks. :)

PR #25, opened 2018-01-01
Changed return methods so some hooks can cancel

Jake-Rich:
OnDispenserGather()
OnCropGather()
OnCollectiblePickup()

Current, it isn't possible to prevent the item given out to the player via these hooks. This allows you to cancel if you return anything other than null.


lukespragg:
Manually merged with today's update.

PR #42, opened 2018-03-01
Fix for downloading failure from github.

Vlad-00003:
To be honest - I have 0 clue what the line ```[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12``` does, but without it - it just don't want to download anything from github wia powershell....

	1>------ Build started: project: Oxide.Rust, Configuration: Debug Any CPU ------
	1>Getting references for public branch of 258550
	1>Determining latest release of DepotDownloader
	1>Could not get DepotDownloader information from GitHub
	1>Request interrupted: Unable to create protected SSL/TLS.
	1>fatal: Not a git repository (or any of the parent directories): .git

[Source of the fix](https://stackoverflow.com/questions/41618766/powershell-invoke-webrequest-fails-with-ssl-tls-secure-channel)

### DIFF
--- a/Steam.ps1
+++ b/Steam.ps1
@@ -81,7 +81,7 @@ function Get-Downloader {
     if (!(Test-Path "$depot_exe") -or ((Get-ChildItem "$depot_exe").CreationTime -gt (Get-Date).AddDays(-7))) {
         # Get latest release info for DepotDownloader
         Write-Host "Determining latest release of DepotDownloader"
-        try {
+        try {[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
             $json = (Invoke-WebRequest "https://api.github.com/repos/SteamRE/DepotDownloader/releases" | ConvertFrom-Json)[0]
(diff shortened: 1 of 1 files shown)


thibmo:
Do you know something about secure socket connections or https or ssl or tls?

What it boils down to, tls12 (TLS1.2) is the kind of encryption used to secure the connection between the client and the server. :)

Vlad-00003:
@thibmo, nope. I know absolutly nothing, as I've never worked with a network in any way. Sadly, but true =(
The thing is - without this code the script doesn't work for me. Checked on 2 different PC and different networks.

First time I've got this then tried to use my serverupdate ps script. Same there. And I'm not the only one - [Click](https://oxidemod.org/threads/github-repository-changes.28181/page-3#post-388420)

lukespragg:
Manually merged in a more global placement for each game extension. Thanks!

Vlad-00003:
Glad I was able to do at least something =)
Feels like I'm a bit usefull now =D

PR #46, opened 2018-03-05
Patch for March 5th update (protocol 2062.158.1)

Ryan-J-D:


lukespragg:
Hook that changed needed to be bumped later, not earlier, else you’d get an NRE with the 2nd arg.

PR #53, opened 2018-04-04
Changed Hook Indexes to support new patcher / Array Pooling

Jake-Rich:
Look at https://github.com/OxideMod/Oxide.Core/pull/37 for details.

Since the MSIL per hook was changed, any hook that applied on a base hook had to be offset (often by -10, though each hook is different)


lukespragg:
Manually merged locally, will be pushed after the weekend and Rust updates. Thanks!

PR #63, opened 2018-04-12
Changed CanExperiment hook.

Vlad-00003:
Changed the CanExperiment hook a bit.
1. Now - it would be called way later, to the point there pendingBlueprint would be chosen already.
2. Added this as a hook argument 
3. Renamed to the OnExperiment to better reflect the changes.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -6483,13 +6483,13 @@
         {
           "Type": "Simple",
           "Hook": {
-            "InjectionIndex": 12,
+            "InjectionIndex": 107,
             "ReturnBehavior": 1,
             "ArgumentBehavior": 4,
-            "ArgumentString": "l0, this.pendingBlueprint",
+            "ArgumentString": "this,l0, this.pendingBlueprint",
             "HookTypeName": "Simple",
-            "Name": "CanExperiment",
-            "HookName": "CanExperiment",
+            "Name": "OnExperiment",
+            "HookName": "OnExperiment",
             "AssemblyName": "Assembly-CSharp.dll",
             "TypeName": "Workbench",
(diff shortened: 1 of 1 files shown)


lukespragg:
This would be better suited as a new hook, not replacing the current.

Vlad-00003:
@lukespragg, I was thinking about that too, but.... The current hook just doesn't make any sense.
As it is in the required function, but it's being called to early to actually have the passed variables inited.
And also - this is impossible to have 2 hooks on the same func atm.

What I need to do - is to prevent some items from being researched. 
So my first idea was simple - use this hook and see if the item is what I wanna block. If so - run the func again. 

But now I can see that this is possible to change the LootSpawn of the workbench on spawn... so I don't think I really need this.

But still - the current hook doesn't make any sense. It pass the variable that is always null, and it doesn't pass self...

Vlad-00003:
Oh, yeah, my bad. 115, actually, would be fine.
Really? For some reason I can't make a new hook in the OxidePatcher (manually mb, but I don't have enought time to learn how the obj format works) - "Hook This Method" button disabled for some reason:
image

lukespragg:
You'd use the "Clone" button on the existing hook.

Vlad-00003:
Wow. Thanks =3
But still - if you  would change this hook to what I've made (name changing is optionall) - both would be resolved - need for the new hook and the current hook arg =)

lukespragg:
The hook change was pushed a little bit ago, with only "this", not "this.pendingBlueprint".

Vlad-00003:
Oh. Yeah. If we got a wb we don't need bp, as it's a public field.... Thanks! =3

PR #74, opened 2018-05-01
Add a hook after Player's metabolism is ran, so we can modify it

Jake-Rich:
The only hook for metabolism is before it is run, so it's really hard to modify it after everything is calculated. This hook allows us to modify bleeding, block sprinting, and also see what values vanilla rust assigned.


lukespragg:
Manually merged as OnPlayerMetabolize.

PR #85, opened 2018-06-09
Added UnityEngine.VehiclesModule to default references

k1lly0u:


lukespragg:
Added along with a couple others people may use.

PR #107, opened 2018-08-25
Ability to change max health of murderer NPC

Jake-Rich:
By default, Rust caps the maximum health of all murderers to 100 and skips the _maxHealth variable it uses for all other entities. This changes behavior so you can set the maximum health and have it work.


lukespragg:
Manually merged, thanks!

PR #108, opened 2018-09-01
Added a OnServerInitialized with Initial startup indicator

austinv900:
Added `private void OnServerInitialized(bool initialStartup);`


lukespragg:
Manually merged with changes (no duplicate hook calls), thanks.

PR #124, opened 2018-12-27
Added in-game team hooks

IvMisticos:


lukespragg:
Manually merged. Keep in mind that these hooks will most likely change when we refactor all hooks across all of the games we support.

PR #126, opened 2019-01-06
Add Extra Overload methods

MisterPixie:
Added 3 new overload methods to allow adding a Icon using a SteamID for PrintToChat & SendReply.


lukespragg:
These methods are all being removed (already have been under develop), so there's no benefit to this, sorry.

PR #136, opened 2019-02-09
OnEntityBuilt Fix

MisterPixie:
Fixes OnEntityBuilt, tested it with BGrade.

Close if i'm wrong, not good at this & could be completely incorrect ahah.


lukespragg:
Good catch with the fix, but bad PR. :P

PR #142, opened 2019-03-16
Add hook OnBigWheelLoss

redBDGR:
Implemented & tested new hook for the BigWheelGame object. This hook will be called from the Payout method inside of BigWheelGame. It will server function for showing when players lose their bets. The hook contains the BigWheelGame object and the gambled item as it's parameters.


lukespragg:
Manually merged, thanks.

PR #144, opened 2019-03-25
Add CanSamSiteShoot

Slydelix:
This gets called before a SAM site attempts to shoot a missile and if a plugin returns non-null value it will not shoot, the target can be obtained from samSite.currentTarget
(Tested, working)


lukespragg:
Assuming you still wanted this merged, so merged manually. :P

PR #145, opened 2019-03-30
Fixed IOnServerUsersSet and IOnServerUsersRemove

Jake-Rich:
Moved IOnServerUsersSet so it is called after the user is removed, as opposed to before it.
Moved IOnServerUsersRemove so it is called right before the user is removed.


lukespragg:
Manually merged into current development branch.

PR #148, opened 2019-04-14
Made playerData visible so offline players can be searched

Jake-Rich:
Right now there is no way to search through offline players for a steam ID. The data was already stored in oxide so I only had to provide a wrapper for it.


lukespragg:
You can already search through offline players with players.All, why would this be any better?

Jake-Rich:
Misread IEnumerable<RustPlayer> as IEnumerable<BasePlayer>

PR #150, opened 2019-05-11
Added OnFlameTurretFire and OnShotgunTrapFire hooks.

gsuberland:
Both of these hooks pass 'this' as the first parameter (FlameTurret or GunTrap) and the target BasePlayer as the second parameter. Returning false prevents the trap from firing. Returning true (or null) allows the trap to fire.

Example plugin code:

```
bool OnFlameTurretFire(FlameTurret turret, BasePlayer player)
{
	Puts("OnFlameTurretFire called! Allowing flame turret to fire.");
	return true;
}

bool OnShotgunTrapFire(GunTrap turret, BasePlayer player)
{
	Puts("OnShotgunTrapFire called! Not allowing shotgun trap to fire.");
	return false; 
}
```

This isn't quite as fully featured as the existing hooks for `AutoTurret`, but it's better than nothing.

These hooks will be of particular use to my [Turret Limits plugin](https://umod.org/plugins/turret-limits) as it will allow me to prevent traps from firing without an active electricity supply in the base.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -1929,6 +1929,54 @@
         {
           "Type": "Simple",
           "Hook": {
+            "InjectionIndex": 94,
+            "ReturnBehavior": 3,
+            "ArgumentBehavior": 4,
+            "ArgumentString": "this, l4 => l2",
+            "HookTypeName": "Simple",
+            "Name": "OnFlameTurretFire",
+            "HookName": "OnFlameTurretFire",
+            "AssemblyName": "Assembly-CSharp.dll",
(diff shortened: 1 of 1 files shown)


lukespragg:
Both of these hooks already exist as CanBeTargeted, but they were disabled due to issues with their associated patches.

gsuberland:
@lukespragg Are there any plans to fix CanBeTargeted to re-enable hooking for these turret types?

lukespragg:
They were broken with the Unity update awhile back, and not sure they can
be fixed with the current patcher.

>

PR #151, opened 2019-05-20
Add hook OnSamSiteTargeted

rostov114:
It is executed immediately after filling in the currentTarget variable in the TargetScan method.
This is a shit attempt to create an analogue of the CanBeTargeted hook for SamSite.
It is more correct to call a hook in the search target, but at the moment it is not possible because of a bug in OxidePatcher.


lukespragg:
Added object OnSamSiteTarget(SamSite, BaseCombatEntity). Thanks!

PR #152, opened 2019-05-21
Modify args from hooks OnExplosiveThrown and OnExplosiveDropped

rostov114:
Adding the third argument to the OnExplosiveThrown and OnExplosiveDropped hooks that will be carried by ThrownWeapon. 
This is done in order to be able to know from which of the entities the BaseEntity was spawned which is passed by the second argument. The work of old plug-ins does not affect because they will ignore the third argument.


lukespragg:
Implemented, thanks!

PR #155, opened 2019-06-30
Added CanAffordToPlace hook

collectVood:
Hello there,
added CanAffordToPlace hook. It executes in CanAffordToPlace.
Returning a non-null value overrides default behavior.
Tested, no problems detected.


rostov114:
Hello!

ArgumentString I think it is necessary to lead to the form "l0, this, a0".
Because at this stage, we already know the player who performs the action.

Thank.

collectVood:
Added BasePlayer to args

lukespragg:
Manually merged, thanks!

PR #156, opened 2019-07-15
New Hooks & Made some methods / fields public

dassjosh:
Added OnEntityEnter and OnEntityLeave on TriggerParent
Added OnSupplyDropLanded on SupplyDrop
Added OnResourceEntityTakeDamage on ResourceEntity
Made FindBurnable public on BaseOven
Made ConsumeFuel public on BaseOven
Made inventory public on ContainerIOEntity

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -8519,6 +8519,110 @@
             "BaseHookName": null,
             "HookCategory": "Vending"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 22,
+            "ReturnBehavior": 0,
+            "ArgumentBehavior": 3,
+            "ArgumentString": "",
+            "HookTypeName": "Simple",
+            "Name": "OnEntityEnter [tirggerparent]",
+            "HookName": "OnEntityEnter",
+            "AssemblyName": "Assembly-CSharp.dll",
+            "TypeName": "TriggerParent",
+            "Flagged": false,
+            "Signature": {
+              "Exposure": 3,
+              "Name": "OnEntityEnter",
+              "ReturnType": "System.Void",
+              "Parameters": [
+                "BaseEntity"
+              ]
+            },
+            "MSILHash": "5IBWr+v3XBXDHyRzNl74W71qn7q01FhGtxVvsYmw2BE=",
+            "BaseHookName": null,
+            "HookCategory": "Entity"
+          }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 24,
+            "ReturnBehavior": 0,
+            "ArgumentBehavior": 3,
+            "ArgumentString": "",
+            "HookTypeName": "Simple",
+            "Name": "OnEntityLeave [triggerparent]",
+            "HookName": "OnEntityLeave",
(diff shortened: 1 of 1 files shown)


lukespragg:
Is the OnResourceEntityKilled hook necessary? Looks to me that the existing OnEntityKill would trigger for that already.

dassjosh:
You're correct it won't be needed. I didn't see that hook was being from OnKilled.

dassjosh:
Looking back at it again I remembered why I added it. It was so you could get access to who killed it from HitInfo since OnEntityKill wouldn't have access to that.

lukespragg:
OnEntithDeath could cover that.

dassjosh:
ResourceEntity inherits from BaseEntity where OnEntityDeath is on BaseCombatEntity so it's not called for ResourceEntities

lukespragg:
BaseCombatEntity is based on BaseEntity, it can be called for either.
- OnEntityDeath(BaseEntity)
- OnEntityDeath(BasePlayer)
- OnEntityDeath(Whatever)

dassjosh:
Yes for any entity that inherits from BaseCombatEntity it will but a ResourceEntity does not inherit from BaseCombatEntity so OnEntityDeath will never be called for it.

lukespragg:
No, you can use the hook as: OnEntityDeath(BaseEntity), both types are based on that.

dassjosh:
Are you saying the name of the hook should be OnEntityDeath instead of OnResourceEntityKilled?

lukespragg:
No, I am saying the existing hook can be used, just change the type to BaseEntity to cover them all, or use the specific type you want such as with the patrol heli.

dassjosh:
Die(HitInfo info) which is on BaseComabatEntity is never called for a ResourceEntity so that hook is never going to be called for it which causes the OnEntityDeath hook to not be called for ResourceEntity. So having OnEntityDeath(ResourceEntity entity) will not be called.

lukespragg:
Ah you're right, I was thinking we had it hooked under another. If we can get it hooked under BaseEntity that would be better, but I don't think there is a method there that provides the HitInfo (looking now).

lukespragg:
So in this case, I'd name the new hook OnEntityDeath as well, then you can either use the base of BaseEntity or the actual type where it is hooked.

dassjosh:
Sounds good to me. I'll make the change. :)

lukespragg:
Don't worry about it, working through all of these today for the update so I have it right in front of me. Thanks though!

lukespragg:
So you feel OnHelicopterStrafeEnter would be better as a post than a pre that is returnable?

dassjosh:
I just wanted it so I could change the number of maxRocket the heli had. I can still do that if you make it Pre with a NextTick so i'm fine either way.

lukespragg:
This is being manually reviewed and merged, so I wouldn't worry about trying to keep up with the other changes.

dassjosh:
I'll close it. I was looking to add more changes for the excavator and try to resolve the IL issue. I'll create a new one once I have those ready.

lukespragg:
Generally hooks should be done so in individual PRs to avoid issues with changes. The IL issue is related to the patcher itself, not so much the hook. We can try the original index that you had for the ResourceEntity hook under OnAttacked, but that comes before the isKilled check.

PR #157, opened 2019-07-17
Player and Structure hooks

rostov114:
CanPayForPlacement - called before attempting to "pay" for the construction.
CanPayForUpgrade - called before attempting to "pay" for upgrade.
OnStabilityCheck - called before trying to check the structure for stability.

Returning a non-null value overrides default behavior.
Tested, no problems detected.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -8519,6 +8519,84 @@
             "BaseHookName": null,
             "HookCategory": "Vending"
           }
+        },
+        {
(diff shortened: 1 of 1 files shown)


lukespragg:
Manually merged with some renames, thanks!

rostov114:
What about hooks?

rostov114:
@lukespragg and hooks will be added?

PR #158, opened 2019-07-23
Vending and Players hooks / Change property in BuildingManager

rostov114:
Hooks:

OnTakeCurrencyItem - Called before attempting to pick up a payment item. Returning a non-null value overrides default behavior.
OnPlayerSetInfo - Called when setting/changing info values for a player (it becomes possible to track for example the inclusion of godmode)

---

Modifiers:

BuildingManager::buildingDictionary change to "public"
BuildingManager::decayEntities change to "public"

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -8519,6 +8519,111 @@
             "BaseHookName": null,
             "HookCategory": "Vending"
           }
+        },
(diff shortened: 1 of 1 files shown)


lukespragg:
Manually merged, thanks!

PR #161, opened 2019-08-09
New hooks for explosives

2chevskii:
Added `object CanBecomeDud(DudTimedExplosive)` hook added. Return non-null to explode anyways.
Added `float SetFuse(TimedExplosive, float)` hook added. Return float to override default fuse length.


2chevskii:
I closed the request since it was made with old broken patcher, before Def updated it. I thought it would be better to make these hooks using the new version, doesn't it?

lukespragg:
No worries, manually merged with some changes, thanks!

PR #162, opened 2019-08-12
Fixes for broken hooks

virtyvoid:
Fixed all previously broken hooks due to the Oxide Patcher issues.
(Of course, a fixed patcher is required: OxideMod/Oxide.Patcher#142)

All fixed hooks been manually tested and results are positive. No exception or crash been observed.
However, for correctness, it is better to double-check all fixed hooks.

Known issues:
- Npc targeting hooks (especially HTN ones) may not work as expected:
Some npcs can still be watching the targets or even attacking them. Need a more stronger amnesia.
- `CanBeTargeted [xxxx] [cleanup]` hooks cannot be viewed in decompiled form.
Supposedly because of a type analysis bug(?) in ICSharpCode.Decompiler library which is seem to be outdated.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -1489,13 +1489,13 @@
             "InjectionIndex": 112,
             "ReturnBehavior": 0,
             "ArgumentBehavior": 4,
-            "ArgumentString": "this, l11",
+            "ArgumentString": "this,r112",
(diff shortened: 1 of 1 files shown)


lukespragg:
Manually merged, thanks!

PR #163, opened 2019-09-04
Add 2 new hooks

Slut:
OnFindSpawnPoint()
  - Return BasePlayer.SpawnPoint to override.
  - Good for minigame servers to override native spawnpoint finding method (removes overhead)

CanPlayerChangeActiveItem(uint newItem)
  - Return non null to cancel


lukespragg:
Manually merged, thanks!

PR #167, opened 2019-09-26
Two new hooks and 2 exposed fields

virtyvoid:
Hooks for I/O puzzle entities:
`object OnCardSwiped(CardReader, BasePlayer, KeyCard)`
Called every time a player uses a KeyCard on a CardReader.
The card reader must be powered.
Returning non-null value cancels the default behavior.
`object OnButtonPress(PressButton, BasePlayer)`
Called every time a player about to push a button.
The button should not be in a pressed state.
Returning non-null value cancels the default behavior.

Fields:
`SpawnGroup::spawnClock`
Exposes internal event scheduler. Can be useful for modifying respawns.
`BaseEntity::triggers`
Exposes the list of all triggers in which the BaseEntity is.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -9112,6 +9112,58 @@
             "BaseHookName": null,
             "HookCategory": "Player"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 54,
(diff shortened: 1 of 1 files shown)


lukespragg:
Manually merged with thanks, thanks!

PR #168, opened 2019-10-01
OnUnloadAmmo hook added

2chevskii:
Even tho ammo unloading is triggered in OnItemAction, it's not really obvious and does not tell if the ammo will be unloaded (if BaseProjectile.canUnloadAmmo is false, then the ammo won't be unloaded, but hook is triggered still).

Return type - object
Arguments - BaseProjectile, Item, BasePlayer


lukespragg:
Manually merged as OnAmmoUnload, thanks.

PR #169, opened 2019-10-03
Real-time player language update

rostov114:
When changing the language through the player menu or through the command "global.language" on the client side, Oxide/uMod accepts the changes immediately.


lukespragg:
Merged, thanks.

PR #171, opened 2019-10-31
New hooks for WireTool

Vlad-00003:
2 new hooks:
```csharp
object OnIOMakeConnection(BasePlayer player, IOEntity input, int inputSlot, IOEntity output, int outputSlot)
{
	Puts("OnIOMakeConnection works!");
	return null;
}
```

Called then player is trying to create connection between two IOEntities. 
At this moment linePoints should be created already, so if you want to prevent connection lines you should set them to null:
- `input.inputs[inputSlot].linePoints = null;`
- `output.outputs[outputSlot].linePoints = null;`

Returning a non-null value cancels default behavior.

```csharp
object OnIORequestClear(BasePlayer player, IOEntity target, int slot, bool isInput)
{
	Puts("OnIORequestClear works!");
	return null;
}
```
Called then player is trying to clear IOEntity slot from existing connection.
isInput defines if the target slot input or ouptut. 
- `IOEntity.IOSlot iOSlot = isInput ? target.inputs[slot] : target.outputs[slot];`

Returning a non-null value cancels default behavior.

I tried to move OnIORequestClear a bit higher, to 87 in case - but even tho oxide patcher doesn't show any errors, as well as dnSpy - server drop clients with `InvalidProgramException: Invalid IL code in WireTool:RequestClear (BaseEntity/RPCMessage): IL_00fa: ret`


Vlad-00003:
Oh, and some off-top while I'm here - what is going on with "OnActiveItemChange" hook?
Why it's up-to-date and deprecated at the same time?
This code sample
```csharp
void OnActiveItemChange(){PrintWarning("Called");}
```
would result in the hook being called 2 times, and even more - since you moved the... original?... hook away, replacing it with `IOnActiveItemChange` - it doesn't have any return behavior anymore.

lukespragg:
> Oh, and some off-top while I'm here - what is going on with "OnActiveItemChange" hook?
> Why it's up-to-date and deprecated at the same time?
> This code sample
> 
> ```cs
> void OnActiveItemChange(){PrintWarning("Called");}
> ```
> 
> would result in the hook being called 2 times, and even more - since you moved the... original?... hook away, replacing it with `IOnActiveItemChange` - it doesn't have any return behavior anymore.

Different signatures.

Vlad-00003:
> > Oh, and some off-top while I'm here - what is going on with "OnActiveItemChange" hook?
> > Why it's up-to-date and deprecated at the same time?
> > This code sample
> > ```cs
> > void OnActiveItemChange(){PrintWarning("Called");}
> > ```
> > 
> > 
> > would result in the hook being called 2 times, and even more - since you moved the... original?... hook away, replacing it with `IOnActiveItemChange` - it doesn't have any return behavior anymore.
> 
> Different signatures.

oh. Right. 
But still - moving it to RustHooks as it is now removes any return behavior.

lukespragg:
Sorry for the delay with this one.

Were you suggesting these be added n the hooks if returning?

input.inputs[inputSlot].linePoints = null;
output.outputs[outputSlot].linePoints = null;

Also, are you using the latest version of the patcher?

Vlad-00003:
> Sorry for the delay with this one.
> 
> Were you suggesting these be added n the hooks if returning?
> 
> input.inputs[inputSlot].linePoints = null;
> output.outputs[outputSlot].linePoints = null;
> 
> Also, are you using the latest version of the patcher?

I was using the one that project automaticly downloads on build.
Setting line points to null? Hm...
This should either be handled by the plugin or the hook can be changed to internal, this doesn't really matter for me.
Also - from what I've found `linePoints` only being stored on the output, so there is no need to clear it on the input.

Vlad-00003:
Well. I did all I could.
Line points will now be handled by RustHooks, and RequestClear hook will now pass the entity that is connected to target.
```csharp
private object OnWireToolConnection(BasePlayer player, IOEntity input, int inputSlot, IOEntity output, int outputSlot)
{
    Puts("OnWireToolConnection works!");
    return null;
}
```
Called then player is trying to create connection between two IOEntities.
Returning a non-null value cancels default behavior.

```csharp
object OnWireTooRequestClear(BasePlayer player, IOEntity target, int slot, bool isInput, IOEntity connectedTo)
{
    Puts("OnWireTooRequestClear works!");
    return null;
}
```
Called then player is trying to clear IOEntity slot from existing connection.
Returning a non-null value cancels default behavior.

lukespragg:
Thanks, merging manually with some minor changes.

PR #175, opened 2019-11-16
Server Side Map Generator Exposures

dassjosh:
Changed the modifiers for the server side rust map generator on staging to public


lukespragg:
These can't be pulled until they make it into a build intended for release. ;)

dassjosh:
Yeah I know it's only on staging branch right now. I just wanted them avaliable once it's merged with main.

PR #178, opened 2019-12-04
New Hook: OnCreateCorpse(BasePlayer, BaseCorpse)

dassjosh:
The hook is called for BasePlayer and all children of BasePlayer after their corpse is created.


lukespragg:
Thanks, merged.

PR #184, opened 2020-01-24
Update Rust.opj

XAuBAH:
please make hook OnSpinWheel return type bool

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -4346,7 +4346,7 @@
           "Type": "Simple",
           "Hook": {
             "InjectionIndex": 16,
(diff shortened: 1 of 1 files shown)


lukespragg:
That isn't really how the patcher works, sorry.

XAuBAH:
> That isn't really how the patcher works, sorry. A return can be added, but it would most likely be object unless wrapped.
Sorry I newbie in commit with github
Please could you do that right way?

lukespragg:
I can add a return to it, yes.

It would be object OnSpinWheel, which you can return true to stop or null to continue.

PR #187, opened 2020-02-19
Add new Hooks and Modifiers

rostov114:
Hooks:

**CanLootEntity** for BaseRidableAnimal - CanLootEntity extension, now it works on horses.
**CanSamSiteTargeted** - It is executed at the moment when SamSite is looking for a target. The logical replacement for the OnSamSiteTarget hook that was created when the OxidePatcher was broken.
**OnCloseConnection** - It is executed when Connection with the player is closed.

Modifiers:
**ResourceEntity::resourceDispenser** - Allows you to edit the dispenser of extracted resources.
**BaseCorpse::resourceDispenser** - Allows you to edit the corpse dispenser.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -10381,6 +10381,81 @@
             "BaseHookName": "OnPlayerSetInfo [server]",
             "HookCategory": "Player"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 26,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
+            "ArgumentString": "l0, this",
+            "HookTypeName": "Simple",
+            "Name": "CanLootEntity [ridable animal]",
+            "HookName": "CanLootEntity",
+            "AssemblyName": "Assembly-CSharp.dll",
+            "TypeName": "BaseRidableAnimal",
+            "Flagged": false,
+            "Signature": {
+              "Exposure": 0,
+              "Name": "RPC_OpenLoot",
+              "ReturnType": "System.Void",
+              "Parameters": [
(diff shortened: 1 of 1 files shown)


lukespragg:
I don't see a point in CanSamSiteTargeted, as it is nearly identical to OnSamSiteTarget with both placement and purpose. There only needs to be one there as a pre.

I'm not sure what OnCloseConnection would be for, as we have OnPlayerDisconnected.

rostov114:
OnSamSiteTarget - Executed when the target is already found.
CanSamSiteTargeted - It is performed in the target search cycle and it is possible to set different types of turrets to behave differently (for example, static turrets will not attack both types of helicopters, but will attack players, and game ones will attack everyone).


OnCloseConnection - It will always be executed when the connection is broken, even if BasePlayer is not created (OnPlayerDisconnected - only worry if BasePlayer is present). It will be convenient to track the reasons for disconnecting players who have not yet managed to log into the server but for some reason have thrown them out (for example, kick from anti-cheat or time out connection).

lukespragg:
Added: void OnClientDisconnect(Connection connection, string reason)

For the others, I'll have to discuss it a bit as I don't think we need both still.

rostov114:
I am in favor of removing OnSamSiteTarget since it runs lower in logic.
CanSamSiteTargeted - in this regard is more universal, it is executed at the moment of scanning targets, like a regular automatic turret.

NWaterbury:
I am also in favor

Vlad-00003:
Oh. Just the moment I was making CanLootEntity hook for BaseRidableAnimal myself.

@lukespragg, is there any chance we may see this in the nearest future? It's been a month :)

lukespragg:
Merged manual with changes for update today.

PR #189, opened 2020-03-09
Added OnBuildingSplit Hook

dassjosh:
void OnBuildingSplit(BuildingManager.Building building, uint newId) - Called when a building is about to be split into new buildings. Called for each building created with the new buildingId


lukespragg:
Thanks, merged manually!

PR #190, opened 2020-03-29
Exposing "internal virtual" methods from TriggerBase

virtyvoid:
This should let modders creating custom triggers based on `TriggerBase` and not use unwanted hooks to monitor players/players collider state.


Slut:
You can create custom triggers by overriding the "OnTriggerEnter" method with the "new" modifier and implementing all of the methods into your trigger manually for the time being :)

lukespragg:
Manually merged for the update today. Also, in your PR, they do not appear to be exposed as public.

PR #191, opened 2020-04-02
Fixing OnVisibilityCheck hook

virtyvoid:
Previous method are no longer in use.
New signature is `uint, string, BaseEntity, BasePlayer, float`.


Vlad-00003:
I guess this one can also be alosed as of 59366a532f8d09a42a06221ff8413e42b77ca1a7

PR #192, opened 2020-04-02
Fix for wire hooks

Vlad-00003:
Fixing wrong signatureIt seems that the functions was slightly changed, so... 
In the OnWireConnect hook Oxide gives us `(BasePlayer, BaseNetworkable, uint, BaseNetworkable, uint)`, where uint is actually just net.id of BaseNetworkables, instead of `(BasePlayer, IOEntity, int, IOEntity, int)`, where int's are slot indexes that are being used.

In the OnWireClear there is a similar problem - `(BasePlayer,  BaseNetworkable, uint, IOEntity.IOSlot, int)`, where uint is acualy nothing but BaseNetworkable net id. And we also lost bool that represents if the slot being cleaned is input or output.

New signatures:

```csharp
private object OnWireConnect(BasePlayer player, IOEntity inputEntity, int inputSlot, IOEntity outputEntity,
	int outputSlot)
{
	Puts("Player {0} trying to connect {1}[{2}] with {3}[{4}]", player, inputEntity, inputSlot, outputEntity,
		outputSlot);
	return null;
}
```

```csharp
private object OnWireClear(BasePlayer player, IOEntity entity, int slot, IOEntity connectedTo, bool isInput)
{
	Puts("Player {0} trying to clear input?({1}) of entity {2}[{3}] that is connected to {4}", player, isInput,
		entity, slot, connectedTo);
	return null;
}
```


Vlad-00003:
Just in case - this is how this hooks looks like now:
OnWireClear
OnWireConnect_init
OnWireConnect_hook

lukespragg:
Thanks for the catch! Merged manually with other fixes.

PR #193, opened 2020-04-03
Added OnDistanceCheck hook into BaseEntity.RPC_Server.MaxDistance

Vlad-00003:
Pretty much same idea as OnVisibilityCheck, it would allow plugin creators to allow\deny interactions no matter how far player is from the entity.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -10080,6 +10080,36 @@
         {
           "Type": "Simple",
           "Hook": {
+            "InjectionIndex": 0,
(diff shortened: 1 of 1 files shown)


Vlad-00003:
```csharp
public object OnDistanceCheck(uint id, string debugName, global::BaseEntity ent, global::BasePlayer player,
	float maximumDistance)
{
	Puts("OnDistanceCheck!");
	return null;
}
```

lukespragg:
Added with changes for next update, thanks.

PR #194, opened 2020-04-03
Added CanChangeSkin hook

Vlad-00003:
Most of the developers when trying to make some unique items - use skins.
So if the player would change skin on the item - the plugin would have no way to check if it's unique or not.

There is another way - storing items uid, but in that case - this items won't work with most of the "backpack" type of plugins, as the original item would get destroyed.

```csharp
private object CanChangeSkin(BasePlayer player, RepairBench bench, int skin, Item item)
{
	Puts("CanChangeSkin works!");
	return null;
}
```

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -3622,6 +3622,32 @@
         {
           "Type": "Simple",
           "Hook": {
+            "InjectionIndex": 22,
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged with changes, thanks.

PR #195, opened 2020-04-04
Moved OnGrowableGathered to earlier index

Vlad-00003:
This would help those who want to change the item BEFORE it was giving to the player.
Example: changing amount.

If we would change amount after the item was given - the player would receive notification that he get original amount, as amount was changed after it was given.

This way - if we would change amount - the notification would show correct value, so we won't have to make another with the difference.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -10261,7 +10261,7 @@
         {
           "Type": "Simple",
           "Hook": {
-            "InjectionIndex": 26,
+            "InjectionIndex": 23,
             "ReturnBehavior": 0,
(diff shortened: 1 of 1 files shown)


lukespragg:
Not ideal, but understandable.

Vlad-00003:
So you would leave it as it is now?

Vlad-00003:
This also makes this hook inconsistent with overs, like OnDispenserGather

Vlad-00003:
Oh, right, I see. 00cfb4310e4a342d25f17fa6f603387dc9f9a9ff

Thanks :3

PR #196, opened 2020-04-05
TriggerParent hooks changed to TriggerComfort

Vlad-00003:
The thing is - from all the triggers available, only TriggerComfort doesn't call the base OnEntity(Leave)Enter method of TriggerBase.

This means that in the current state - OnEntityEnter won't be called if the player entered TriggerComfort, and would be called twice on TriggerParent.
image
image
image


lukespragg:
Thanks, merged.

PR #197, opened 2020-04-16
Handling zero index in FindPlayers

Vlad-00003:
```csharp
"vlad-00003".IndexOf("vlad",StringComparison.OrdinalIgnoreCase) == 0
```
Pretty much self-explaining, it also worked that way before 7396a0968b52d261a59ca9eb1f2daed3452ee674


lukespragg:
Thanks for the catch! I'll need to apply it across all games though.

PR #205, opened 2020-07-10
Expose various SmartAlarm fields for plugins

haggbart:
Expose fields: 

NotificationList _subscriptions 
string _notificationTitle
string _notificationBody
float _lastSentTime


lukespragg:
Merged, thanks.

PR #207, opened 2020-07-30
Fix OnPlayerAttack [Projectile]

features-not-bugs:
Rust released an update ~9 hours ago (build id: 5346990), this began causing IL errors during patching
ilerror

This was causing players to disconnect due to RPC errors on creation of any projectile.

Pre fix:
pre-fix

Post fix:
post-fix

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -582,7 +582,7 @@
         {
           "Type": "Simple",
           "Hook": {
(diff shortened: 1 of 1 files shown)


lukespragg:
Didn't get the notification for this update, but it has been updated and will be released shortly. Thanks!

PR #212, opened 2020-08-10
Add RF signal hooks and expose ModularCarGarage fields

WheteThunger:
The following RF signal hooks were added.
- Add OnAddRFListener(int frequency, IRFObject obj)
- Add OnRemoveRFListener(int frequency, IRFObject obj)
- Add OnAddRFBroadcaster(int frequency, IRFObject obj)
- Add OnRemoveRFBroadcaster(int frequency, IRFObject obj)

Various ModularCarGarage methods/properties/fields were made public.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -11673,6 +11673,114 @@
             "BaseHookName": "OnDefaultItemsReceive",
             "HookCategory": "Player"
           }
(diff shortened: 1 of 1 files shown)


lukespragg:
Manually merged with changes, thanks!

PR #213, opened 2020-08-10
Created OnDropBonusItems Hook

dassjosh:
void OnDropBonusItems(Baseplayer player, Item item)
Called when bonus items are dropped from a loot container


lukespragg:
Manually merged with changes, thanks!

PR #214, opened 2020-08-11
Added CanHit hook 

Wolfleader101:
Added CanHit hook to allow modifications to what hammers can hit

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -14,14 +14,14 @@
             "HookName": "InitOxide",
             "AssemblyName": "Assembly-CSharp.dll",
             "TypeName": "Bootstrap",
-            "Flagged": false,
+            "Flagged": true,
             "Signature": {
               "Exposure": 2,
               "Name": "Init_Tier0",
               "ReturnType": "System.Void",
               "Parameters": []
             },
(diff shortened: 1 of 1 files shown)


rostov114:
CanHit is client side method. In server is not using.

rostov114:
Method found in

image

Found 1 match in BaseMelee
image

Method not calling in server.

-----

Client (version 2183)

Found 2 match in BaseMelee
image

rostov114:
@lukespragg can close PR?
Hook not work.

Wolfleader101:
well yes its not working bc the patch screwed it up a little 🤣  i will fix it later. tried last night when free and it didnt work

Slut:
Hook will work correctly, it would just need to be hooked in the override (Hammer).

Make a clean server install (ensure you're on the right branch and fully updated), install oxide then update server again to remove patched files then load the latest opj from github and make your changes.

One thing I would recommend (but not entirely sure) is to remove all the "_Original" files

Wolfleader101:
> Hook will work correctly, it would just need to be hooked in the override (Hammer).
> 
> Make a clean server install (ensure you're on the right branch and fully updated), install oxide then update server again to remove patched files then load the latest opj from github and make your changes.
> 
> One thing I would recommend (but not entirely sure) is to remove all the "_Original" files

ok thanks

Wolfleader101:
followed @Slut and it still is changing lots of lines in opj file

PR #218, opened 2020-08-20
Update CanTakeCutting and add OnRemoveDying for GrowableEntity

rostov114:
CanTakeCutting - logical moving. Execute after checks.
OnRemoveDying - execute before remove dying GrowableEntity


lukespragg:
Thanks, manually merged with minor change!

PR #219, opened 2020-08-28
Added OnItemRefill Hook

dassjosh:
Added hook object OnItemRefill(BasePlayer player, Item item)

Tested and working


lukespragg:
Added with changes, thanks!

PR #221, opened 2020-09-01
Add OnFrequency hooks

MrBlue:
Add OnFrequencyChange [Broadcaster] hook
Add OnFrequencyChange [Receiver] hook
Add OnFrequencyChange [Detonator] hook
Add OnFrequencyChange [PagerEntity] hook

```csharp
private object OnFrequencyChange(IRFObject item, int frequency, BasePlayer player)
{
    Puts("OnFrequencyChange works!");
    return null;
}
```

Add OnFrequencyChanged [Broadcaster] hook
Add OnFrequencyChanged [Receiver] hook
Add OnFrequencyChanged [Detonator] hook
Add OnFrequencyChanged [PagerEntity] hook
```csharp
private void OnFrequencyChanged(IRFObject item, int frequency, BasePlayer player)
{
    Puts("OnFrequencyChanged works!");
}
```

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -11673,6 +11673,214 @@
             "BaseHookName": "OnDefaultItemsReceive",
             "HookCategory": "Player"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 35,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
+            "ArgumentString": "this, l0, a0.player",
+            "HookTypeName": "Simple",
+            "Name": "OnFrequencyChange [Broadcaster]",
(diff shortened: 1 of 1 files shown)


WheteThunger:
Hmm, this is the 3rd PR this month for adding hooks to RF receivers/broadcasters. I wonder what everyone is working on to suddenly want these.

Curious if you can achieve what you're looking for with hooks in #212. I added them primarily so I could detect the frequency changing on an RF receiver. Main difference is there's isn't a player so your approach might be more widely useful.

MrBlue:
We needed a way to create a few reserved frequencies, this way we can check if a player is using one of our reserved frequencies and block them from using it.

lukespragg:
Merged, thanks!

PR #222, opened 2020-09-01
Add Connection argument to OnPlayerSpawn hook

MrBlue:
**Passes Connection as second argument on the OnPlayerSpawn hook**

Before:
```csharp
object OnPlayerSpawn(BasePlayer player)
{
    Puts("OnPlayerSpawn works!");
    return null;
}
```
After:
```csharp
object OnPlayerSpawn(BasePlayer player, Connection connection)
{
    Puts("OnPlayerSpawn works!");
    return null;
}
```

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -2727,7 +2727,7 @@
             "InjectionIndex": 12,
             "ReturnBehavior": 1,
             "ArgumentBehavior": 4,
-            "ArgumentString": "l1",
+            "ArgumentString": "l1, a0",
             "HookTypeName": "Simple",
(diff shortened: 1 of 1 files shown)


lukespragg:
Why is this necessary when the game already has this available via BasePlayer? If you expose it here, then it wouldn't make sense to not have it for every other BasePlayer hook too.

MrBlue:
Connection is not set at that point because the hook is very early in the connection process.

lukespragg:
Re-wrote hook to combine patch and base, and set Connection before hook is called.

PR #226, opened 2020-09-06
Add CanMoveVehicleModule hook

WheteThunger:
This hook allows overriding whether a vehicle module can be removed from a car at a lift.

I was thinking this might be an issue for performance since it's called in `VehicleFixedUpdate()`, but it seems to be only when the vehicle is in an editable location (on a car lift), and only at most once per second per module. If that's too often, I can alternatively expose `BaseVehicleModule.timeSinceItemLockRefresh` and override it to a big negative number in a plugin to prevent the game from overriding the locked state for a specific module after I've locked/unlocked it.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -12160,6 +12160,30 @@
             "BaseHookName": null,
             "HookCategory": "Resource"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 3,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 1,
+            "ArgumentString": null,
+            "HookTypeName": "Simple",
+            "Name": "CanMoveVehicleModule",
+            "HookName": "CanMoveVehicleModule",
+            "AssemblyName": "Assembly-CSharp.dll",
+            "TypeName": "BaseVehicleModule",
+            "Flagged": false,
+            "Signature": {
+              "Exposure": 2,
+              "Name": "CanBeMovedNow",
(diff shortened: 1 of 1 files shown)


lukespragg:
Thanks, merged with changes.

WheteThunger:
A couple comments about the changes you made.

- In terms of naming, this hook is not called when a vehicle module is moved, so the name "OnVehicleModuleMove" may be misleading. It's called in advance to determine whether to lock or unlock a vehicle module.
- OnVehicleModuleMove only allows returning true, so I can only *allow* a vehicle module to be moved, but I cannot **disallow** it, which was actually the original purpose of this hook. I want to both be able to unlock an engine module that contains parts, as well as block a non-storage module from being moved (because it has a turret or other deployable attached).

lukespragg:
Move is pre, Moved is post; this is the naming we've been following rather than having Can hooks.

The other is an oversight, so I'll take a look at it again.

WheteThunger:
In terms of naming again, this hook is not called when a player tries to move a vehicle module. Whether a player can do that is determined by whether the module item is locked. The locking state is changed automatically by the hooked method, both periodically (once per second while at a lift), and when an item enters/exits the module's storage container (for VehicleModuleStoage). Given all of this, it might be better to name it something like `OnVehicleModuleLock`.

lukespragg:
Right, but we are moving away from Can hooks, but I based it on the name of the hook in the PR, which would also imply a module being moved, not locked/unlocked. If another name sounds better, we can go with that, just let me know.

PR #231, opened 2020-09-11
Add OnLootEntityEnd for DroppedItemContainer

rostov114:
Tested, work.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -12602,6 +12602,32 @@
             "BaseHookName": null,
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 0,
+            "ReturnBehavior": 0,
+            "ArgumentBehavior": 4,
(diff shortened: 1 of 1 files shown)


rostov114:
@lukespragg  

For reflection. Can all OnLootEntityEnd - delete and make one?
Add to RustHooks.cs by handling OnPlayerLootEnd and if there is entitySource not null calli OnLootEntityEnd?

rostov114:
Something like that. On the plus side, it catches all loot endings, even for which there are no hooks.

Of the minuses. Plugins may break.

image

```csharp	
[HookMethod("OnPlayerLootEnd")]
private void OnPlayerLootEnd(PlayerLoot inventory)
{
    if (inventory.entitySource)
    {
        Interface.CallHook("OnLootEntityEnd", inventory.GetComponent<BasePlayer>(), inventory.entitySource);
    }
}

lukespragg:
The change is something that could be considered, but unsure we want to do that right now. I'll likely merge the other PR that has the additional hook location too.

PR #232, opened 2020-09-16
Add OnSwitchToggled hook for ElectricSwitch

WheteThunger:
```csharp
void OnSwitchToggled(ElectricSwitch switch, BasePlayer player)
```

This adds a post-hook variant of OnSwitchToggle called OnSwitchToggled that runs after the switch is toggled successfully. Various plugins using OnSwitchToggle should probably change to use OnSwitchToggled eventually.


lukespragg:
Thanks, merged manually!

PR #233, opened 2020-09-17
Add OnTurretRotate hook

WheteThunger:
```csharp
object OnTurretRotate(AutoTurret turret, BasePlayer player)
```

Returning non-null prevents the default behavior. Planning to use this to rotate at a custom angle instead of 180 degrees.


WheteThunger:
Closing since this was merged into develop.

PR #234, opened 2020-09-18
Add OnItemLock and OnItemUnlock hooks

WheteThunger:
```csharp
object OnItemLock(Item item)
```
```csharp
object OnItemUnlock(Item item)
```

Returning non-null will prevent the default behavior.

Decompiled code:

```csharp
public void LockUnlock(bool bNewState)
{
    if (this.HasFlag(Item.Flag.IsLocked) == bNewState)
    {
        return;
    }
    if (bNewState && Interface.CallHook("OnItemLock", this) != null)
    {
        return;
    }
    if (!bNewState && Interface.CallHook("OnItemUnlock", this) != null)
    {
        return;
    }
    this.SetFlag(Item.Flag.IsLocked, bNewState);
    this.MarkDirty();
}
```

This is intended to be used to prevent locking or unlocking a vehicle module item. The vehicle module can be obtained from the item with roughly `item.parent.entityOwner.GetModuleForItem(item)`.

It turns out whether an item is locked is just a visual indicator, so I additionally opened #235 to allow blocking or forcibly allowing a vehicle module to be moved.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -12602,6 +12602,138 @@
             "BaseHookName": null,
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Modify",
+          "Hook": {
+            "InjectionIndex": 6,
+            "RemoveCount": 0,
+            "Instructions": [
+              {
+                "OpCode": "ldarg_1",
+                "OpType": "None",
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged, thanks!

PR #235, opened 2020-09-18
Add OnVehicleModuleMove hook

WheteThunger:
```csharp
object OnVehicleModuleMove(ModularCar car, BasePlayer player, Item item)
```

Called when a player tries to move or drop a vehicle module item that's in a vehicle's inventory.

Note: Allowing both `true` and `false` return values is intentional to allow the use case of removing a module even when it has items in its storage container, as well as to prevent a module from being moved if it has entity attachments such as turrets.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -12602,6 +12602,33 @@
             "BaseHookName": null,
             "HookCategory": "Entity"
           }
+        },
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged with changes, thanks!

PR #236, opened 2020-09-26
Add MiniCopter OnEngineStartup hook

WheteThunger:
Signature:
```csharp
object OnEngineStartup(MiniCopter minicopter)
```

Decompiled code:
```csharp
public void EngineStartup()
{
    if (Interface.CallHook("OnEngineStartup", this) == null && !Waterlogged())
    {
        Invoke(EngineOn, 5f);
        SetFlag(Flags.Reserved4, b: true);
    }
}
```

Edit: Not sure how long this is going to last given that FP is changing how engines work for cars and minicopters due to the introduction of work carts.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -12602,6 +12602,30 @@
             "BaseHookName": null,
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Simple",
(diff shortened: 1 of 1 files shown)


lukespragg:
Manually merged with changes, thanks.

PR #239, opened 2020-09-27
Add hooks for toggling hot air balloons

WheteThunger:
```csharp
object OnHotAirBalloonToggle(HotAirBalloon hab, BasePlayer player)
```
```csharp
void OnHotAirBalloonToggled(HotAirBalloon hab, BasePlayer player)
```

Decompiled code:
```csharp
public void EngineSwitch(RPCMessage msg)
{
    if (Interface.CallHook("OnHotAirBalloonToggle", this, msg.player) == null)
    {
        bool b = msg.read.Bit();
        SetFlag(Flags.On, b);
        if (IsOn())
        {
            Invoke(ScheduleOff, 60f);
            Interface.CallHook("OnHotAirBalloonToggled", this, msg.player);
        }
        else
        {
            CancelInvoke(ScheduleOff);
            Interface.CallHook("OnHotAirBalloonToggled", this, msg.player);
        }
    }
}
```

I wrote it this way (duplicated the hook call) because I couldn't easily move the hook call to the end of the inside of the `if {}` block.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -12602,6 +12602,84 @@
             "BaseHookName": null,
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 0,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
+            "ArgumentString": "this, a0.player",
+            "HookTypeName": "Simple",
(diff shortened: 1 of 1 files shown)


WheteThunger:
I tried improving this with some patch hooks but it wasn't very straight forward. Probably makes more sense to just duplicate the hook call like I've done.

PR #242, opened 2020-09-28
Add hook for assigning car modules

WheteThunger:
```csharp
object OnCarModulesAssign(ModularCar car)
```

Overriding this by returning non-null will allow plugins to replace the default module presets with custom ones.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -12602,6 +12602,30 @@
             "BaseHookName": null,
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 11,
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged wit hchanges, thanks!

WheteThunger:
Looked at the commit:
https://github.com/OxideMod/Oxide.Rust/commit/8c4341317d86ab88cfbdd372d440a7c76f0ead6c

The name `OnVehicleModulesSpawn(ed)` is misleading because the vehicle modules entitie don't "spawn" until later (some time later, couldn't find the logic determining this). This method simply adds the vehicle module items to the car's module inventory.

lukespragg:
Okay, renamed.

PR #243, opened 2020-09-28
Add hook and expose fields for engine stats

WheteThunger:
```csharp
object OnEngineStatsRefresh(VehicleModuleEngine engineModule, EngineStorage engineStorage)
```

Overriding this by returning non-null will allow plugins to set engine stats arbitrarily (using the fields exposed in this PR), avoiding the need for actual engine parts. Now I can stop wasting time making plugins related to engine parts.


lukespragg:
Merged with addition, thanks!

PR #246, opened 2020-10-06
Removing OnEntityKill [BradleyAPC] as it's obsolete

Vlad-00003:
I'm not sure why it was created in the first place, but now - it calls OnEntityKill twice on the bradley
BradleyAPC.OnKilled calls BaseCombatEntity.OnKilled, that calls BaseNetworkable.Kill where OnEntityKill already presented.

This way the hook would be called twice - before actuall object being destroyed and after: 
image
image
image
image

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -9612,32 +9612,6 @@
         {
           "Type": "Simple",
           "Hook": {
-            "InjectionIndex": 4,
-            "ReturnBehavior": 1,
-            "ArgumentBehavior": 1,
-            "ArgumentString": null,
-            "HookTypeName": "Simple",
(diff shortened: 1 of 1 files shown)


soulfly006:
basehelicopter also ?

Vlad-00003:
Ok, I admit it - removing the hook entirly might not be the best idea, renaming probably? 
As the one in the bradley is called before any actions are taking on the class, and the other one is after.

And yeah - heli works the same way, also called twice.

lukespragg:
I imagine a rename would be best.

Vlad-00003:
Hm. It's not about entity, it's just heli and bradley
No idea tbh

PR #249, opened 2020-10-22
ItemModCookable hooks

Vlad-00003:
private object OnItemCook(Item item, float delta) - being called at the start of the cycle. Usefull to prevent some items from being cooked in that specific furnace
private object OnItemCooked(Item item, float delta) - being called right before spawning becomeOnCooked, usefull to change the result of cooking

Both hooks are usefull in the situation where you need to modify either speed (canceling cycle in the first hook, doing everything in the plugin) or outcome (creating items with skin as a result of burning)
The only way to do this now is to re-create cooking functions in the plugin, and makes it impossible to have 2 plugins doing simular things at once

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -13056,6 +13056,60 @@
             "BaseHookName": "OnCounterModeToggle",
             "HookCategory": "Entity"
           }
+        },		
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 0,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
+            "ArgumentString": "a0,a1",
+            "HookTypeName": "Simple",
+            "Name": "OnItemCook",
+            "HookName": "OnItemCook",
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged with some changes.

lukespragg:
@MrBlue ran into an issue with the OnItemCook placement, so holding off on these for now.

lukespragg:
OnOvenCook was added recently, which should handle this. Thanks!

Vlad-00003:
> OnOvenCook was added recently, which should handle this. Thanks!

No, it won't...
Accepted hook can not handle cookable item, it only works on the oven itself, having no way to modify anything related to the cooked item itself

Sure, it is possible to interrupt OnOvenCook and iterrating over oven inventory, but it's not the same thing
ItemModCookable working OnCycle in the game, not in the Cook method of the oven

Vlad-00003:
Or you actually didn't notice that OnOvenCooked provides fuel (wood, lowgrade, etc) as an item? 
And the slot... I'm not even sure what is it but it's null for both furnaces

PR #252, opened 2020-10-26
Temporarily disable minicopter OnEngineStart(ed) hooks

WheteThunger:
These hooks are currently kicking players when starting minicopters.

Also included other changes from the `release` branch.


WheteThunger:
Added commit to remove all OnEngineStart(ed) hooks since it's affecting boats as well, and flagging wasn't fixing minicopters.

PR #255, opened 2020-10-31
Add OnEngineLoadoutRefresh hook and expose fields

WheteThunger:
```csharp
object OnEngineLoadoutRefresh(EngineStorage engineStorage)
```

This is a more precise hook than `OnEngineStatsRefresh`, which will be used specifically for counting engine parts and determining the overall engine stats that result. Doesn't need a post hook.

Also exposed several fields/methods as public on `VehicleModuleEngine` and `EngineStorage`.


lukespragg:
Merged.

PR #264, opened 2020-11-23
Added OnOvenCook(ed) hooks

0x89A:
- Added object OnOvenCook(BaseOven oven, Item burnable)
- Added void OnOvenCooked(BaseOven oven, Item burnable, BaseEntity slot)

Called every 0.5 seconds while the furnace is cooking.


jbrazio:
Could the `OnOvenConsumeFuel` hook be implemented as I suggested on the main thread about this ?

PR #268, opened 2020-12-14
Add OnIORefCleared hook

WheteThunger:
```csharp
void OnIORefCleared(IOEntity.IORef ioRef, IOEntity ioEntity)
```

Called when a wire has been cleared from an IO entity (usually two entities at once, so this will be called twice). This can happen a number of different ways, such as when cleared via a wire tool, or when either entity is killed or moves.

The following example plugin uses this hook to detect when a wire is cleared, in order to send power in its place. Ideally the hook could provide more information but it would have to be implemented in several others places, and this was sufficient for my use case.
https://github.com/WheteThunger/PowerlessElectronics/blob/master/PowerlessElectronics.cs

```csharp
public void Clear()
{
    global::IOEntity OxideGen_0 = this.ioEnt;
    this.ioEnt = null;
    this.entityRef.Set(null);
    Interface.CallHook("OnIORefCleared", this, OxideGen_0);
}
```

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -14626,6 +14626,69 @@
             "BaseHookName": "OnFuelConsume",
             "HookCategory": "Item"
           }
+        },
+        {
+          "Type": "Modify",
+          "Hook": {
+            "InjectionIndex": 0,
+            "RemoveCount": 0,
+            "Instructions": [
+              {
+                "OpCode": "ldarg_0",
+                "OpType": "None",
+                "Operand": null
+              },
+              {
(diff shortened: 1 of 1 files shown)


Vlad-00003:
Are you sure that OnWireClear(BasePlayer player, IOEntity entity, int slot, IOEntity connectedTo, bool isInput) from  #171 won't do the trick?

WheteThunger:
Yes, I tried using that one originally, but it only runs when a player clears it, not for the other cases.

lukespragg:
Merged, thanks!

PR #269, opened 2020-12-15
Add tech tree hooks

WheteThunger:
```csharp
object CanPlayerUnlock(BasePlayer player, TechTreeData.NodeInstance node, TechTreeData techTree)
```

- Called when a player tries to unlock an item in a tech tree (even when the button says "no path")
- Returning true or false will allow or disallow the unlock, null will result in the default behavior
- Useful for disallowing an item from being unlocked, such as globally or based on permissions

```csharp
object OnUnlockPathCheck(BasePlayer player, TechTreeData.NodeInstance node, TechTreeData techTree)
```

- Called after the `CanPlayerUnlock` check, when determining if the player has unlocked the prerequisite items
- Returning true or false will allow or disallow the unlock, null will result in the default behavior
- Useful for customizing prerequisites without conflicting with CanPlayerUnlock

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -14626,6 +14626,60 @@
             "BaseHookName": "OnFuelConsume",
             "HookCategory": "Item"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 0,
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged with some changes, thanks!

PR #270, opened 2020-12-18
Exposed NeonSign and Signage members and functions

0x89A:
Exposed:
- float NeonSign.animationSpeed
- int NeonSign.currentFrame
- List<NeonSign.Lights> NeonSign.frameLighting
- bool NeonSign.isAnimating
- Action NeonSign.animationLoopAction
- void NeonSign.EnsureInitialized()
- void Signage.EnsureInitialized()

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -27850,6 +27850,139 @@
             "Parameters": []
           },
           "MSILHash": ""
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged, thanks!

PR #272, opened 2020-12-23
Add OnInventorySaved hook

WheteThunger:
```csharp
void OnInventorySaved(PlayerInventory inventory, ItemContainer container, PlayerInventory.Type inventoryType, ProtoBuf.UpdateItemContainer updateItemContainer, bool sendToEveryone)
```

- Called when sending an updated player inventory to one or more clients, such as when the player moves items around in their inventory
- No return behavior

This is similar to OnEntitySaved, in that an Item was just saved into a protobuf object and about to be sent over the network, except in this case the networking is happening via RPC so OnEntitySaved isn't called, and there's no distinct entity (player inventories and items do not derive from BaseEntity or BaseNetworkable).

Example usage: Trick the client into thinking it has wood in its inventory (in a hidden slot). Note: To fully trick the client in all cases, you would also need to hook OnEntitySaved for BasePlayer, but excluding that here for brevity.

```csharp
Item wood;
void OnServerInitialized()
{
    wood = ItemManager.CreateByName("wood", 10000);
    wood.position = 24;
}

void OnInventorySaved(PlayerInventory inventory, ItemContainer container, PlayerInventory.Type inventoryType, ProtoBuf.UpdateItemContainer updateItemContainer, bool sendToEveryone)
{
    if (inventoryType == PlayerInventory.Type.Main)
    {
        updateItemContainer.container[0].contents.Add(wood.Save());
    }
}
```

```csharp
public void SendUpdatedInventory(global::PlayerInventory.Type type, global::ItemContainer container, bool bSendInventoryToEveryone = false)
{
    using (UpdateItemContainer updateItemContainer = Pool.Get<UpdateItemContainer>())
    {
        updateItemContainer.type = (int)type;
        if (container != null)
        {
            container.dirty = false;
            updateItemContainer.container = Pool.Get<List<ProtoBuf.ItemContainer>>();
            updateItemContainer.container.Add(container.Save());
        }
        Interface.CallHook("OnInventorySaved", this, container, type, updateItemContainer, bSendInventoryToEveryone);
        if (bSendInventoryToEveryone)
        {
            base.baseEntity.ClientRPC<UpdateItemContainer>(null, "UpdatedItemContainer", updateItemContainer);
        }
        else
        {
            base.baseEntity.ClientRPCPlayer<UpdateItemContainer>(null, base.baseEntity, "UpdatedItemContainer", updateItemContainer);
        }
    }
}
```

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -14626,6 +14626,69 @@
             "BaseHookName": "OnFuelConsume",
             "HookCategory": "Item"
           }
+        },
+        {
+          "Type": "Modify",
+          "Hook": {
+            "InjectionIndex": 0,
+            "RemoveCount": 0,
+            "Instructions": [
+              {
+                "OpCode": "ldarg_0",
+                "OpType": "None",
+                "Operand": null
+              },
+              {
+                "OpCode": "ldfld",
+                "OpType": "Field",
+                "Operand": "Assembly-CSharp|IOEntity/IORef|ioEnt"
+              },
+              {
+                "OpCode": "stloc_0",
+                "OpType": "None",
+                "Operand": null
+              }
+            ],
+            "HookTypeName": "Modify",
+            "Name": "OnIORefCleared [patch]",
+            "HookName": "OnIORefCleared [patch]",
+            "AssemblyName": "Assembly-CSharp.dll",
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged with changes, thanks!

PR #273, opened 2020-12-24
Expose car physics method/fields as public

WheteThunger:
Exposing `ModularCarPhysics.GetModifiedDrag()` and `ModularCarPhysics.timeSinceWaterCheck` so I can overwrite the existing drag computation logic for underwater driving.


lukespragg:
Merged, thanks!

PR #274, opened 2020-12-27
Add OnResearchCostDetermine hook

WheteThunger:
- int OnResearchCostDetermine(ResearchTable table, Item item)
- int OnResearchCostDetermine(ItemDefinition itemDefinition)

Deprecate OnItemScrap(ResearchTable table, Item item)


lukespragg:
Merged with changes as discussed, thanks!

PR #276, opened 2020-12-30
Add OnEntityFlagsNetworkUpdate hook

WheteThunger:
### OnEntityFlagsNetworkUpdate

- Called when an entity is sending a network update with just BaseEntity flags
- Returning a non-null value overrides default behavior

```csharp
object OnEntityFlagsNetworkUpdate(BaseEntity entity)
{
    Puts("OnEntityFlagsNetworkUpdate works!");
    return null;
}
```

Example in https://github.com/WheteThunger/LootablePoweredTurrets

```csharp
protected void SendNetworkUpdate_Flags()
{
    if (Rust.Application.isLoading || Rust.Application.isLoadingSave || base.IsDestroyed || net == null || !isSpawned)
    {
        return;
    }
    using (TimeWarning.New("SendNetworkUpdate_Flags"))
    {
        LogEntry(LogEntryType.Network, 2, "SendNetworkUpdate_Flags");
        if (Interface.CallHook("OnEntityFlagsNetworkUpdate", this) == null)
        {
            List<Connection> subscribers = GetSubscribers();
            if (subscribers != null && subscribers.Count > 0 && Network.Net.sv.write.Start())
            {
                Network.Net.sv.write.PacketID(Message.Type.EntityFlags);
                Network.Net.sv.write.EntityID(net.ID);
                Network.Net.sv.write.Int32((int)flags);
                SendInfo info = new SendInfo(subscribers);
                Network.Net.sv.write.Send(info);
            }
            OnSendNetworkUpdateEx.SendOnSendNetworkUpdate(base.gameObject, this);
        }
    }
}
```

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -14896,6 +14896,30 @@
             "BaseHookName": "OnIORefCleared [patch]",
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 27,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 1,
+            "ArgumentString": null,
+            "HookTypeName": "Simple",
+            "Name": "OnEntityFlagsNetworkUpdate",
+            "HookName": "OnEntityFlagsNetworkUpdate",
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged, thanks!

PR #278, opened 2021-01-07
Removing static modifier from EntityComponent.baseEntity

Vlad-00003:
I really don't know why we have static modifier for a non-static getter in the class

I also don't know why this doesn't cause any other problems, but the thing is - I have a plugin that contains custom class inherited from EntityComponent. Now every time I'm trying to use it I get IL error due to the fact that baseEntity can not be used in non-static context... while it is non-static property

image

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -28981,8 +28981,7 @@
           "TypeName": "EntityComponent`1",
           "Type": 2,
           "TargetExposure": [
-            2,
-            4
+            2
           ],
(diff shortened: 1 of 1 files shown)


lukespragg:
Per the discussion in the other PR, this is a rare patcher bug where it automatically assigns static in some cases, unsure why.

Vlad-00003:
Oh... I see

But the point stands - we have to revert this (:

PR #280, opened 2021-02-03
Add missing expiry arguments for user banned hooks

jonaslagoni:
Checked the hook configuration `Interface.CallHook("IOnServerUsersSet", uid, group, username, notes, expiry);` which already provides the arguments just needs to be passed along. Let me know if anything else needs to change 👍


lukespragg:
Thanks, merged!

PR #279, opened 2021-02-03
Update Rust.opj add OnPlayerDisconnected

XAuBAH:
please could you add OnPlayerDisconnected to
ServerMgr.ReadDisconnectReason(Message packet)
this disconnections is not caching now

PR #283, opened 2021-02-11
Add hook OnClientDisconnectReason

rostov114:
Executing hook on disconnect player with reason.

```csharp
void OnClientDisconnectReason(Network.Connection connection, string reason)
```


rostov114:
This hook has been requested in PR #279

lukespragg:
Does the existing OnClientDisconnect hook not catch this already?

rostov114:
No.

OnClientDisconnect - runs at the level of the Network library. In it, you can track when a person flew out on Time Out.
OnClientDisconnectReason - processes the `Message.Type.DisconnectReason` packet from the client.

They complement each other.

lukespragg:
Would it make sense to have them both named the same then?

rostov114:
I named it from the logic that it is the client / connection that is disconnected.
But I will not mind if you rename it.

lukespragg:
Right, I was going to rename the other to OnClientDisconnected and have this one as OnClientDisconnected as well.

rostov114:
Probably more correct this way:

OnClientDisconnect(Connection, string) -> OnClientDisconnected(Connection, string)   ( since it is executed last )
OnClientDisconnectReason(Connection, string) -> OnClientDisconnect(Connection, string)

lukespragg:
Thanks, merged with the suggestions.

PR #284, opened 2021-02-21
Added OnPlayerAddModifiers Hook

dassjosh:
Added object OnPlayerAddModifiers(BasePlayer player, Item item, ItemModConsumable consumable)


lukespragg:
Thanks, merged!

PR #285, opened 2021-02-24
Add OnNpcConversationEnded hook

WheteThunger:
```csharp
void OnNpcConversationEnded(NPCTalking npcTalking, BasePlayer player)
```
Called when a player select a dialog option that closes the NPC conversation, or when the player clicks the X button to close it.


lukespragg:
Thanks, merged!

PR #294, opened 2021-04-05
OnStructureUpgrade changes & added OnItemStacked

birthdates:
Injecting this method after the BaseCombatEntity#SecondsSinceAttacked check doesn't make any sense, seen as OnStructureRepair does it before.

**Before Commit**
image

**After Commit**
image

**Current OnStructureRepair**
image

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -637,7 +637,7 @@
         {
           "Type": "Simple",
           "Hook": {
-            "InjectionIndex": 37,
+            "InjectionIndex": 33,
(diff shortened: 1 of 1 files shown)


lukespragg:
Please make sure to use separate branches for each PR/change. It's unlikely to be merged directly if there are a bunch of things globed together.

birthdates:
> Please make sure to use separate branches for each PR/change. It's unlikely to be merged directly if there are a bunch of things lobbed together.

I did not realize this was getting pushed here lol. Will make a new one

PR #296, opened 2021-04-11
Add OnItemStacked hook

birthdates:
I've added the hook "OnItemStacked" hook because there's no way of determining if CanStack was successful.

Example:
```csharp
void OnItemStacked(Item newItem, Item oldItem, ItemContainer newContainer)
```

**Code**

image


lukespragg:
Squashed and merged, thanks!
- void OnItemStacked(Item itemBeingMoved, Item destinationItem, ItemContainer destinationContainer)

PR #302, opened 2021-04-26
Added OnConvertWater Hook

BadMandarin:
Added object OnConvertWater(WaterPurifier waterPurifier, float timeCooked)


lukespragg:
Merged, but with a name change and additional post hook. Thanks!
- object OnWaterPurify(WaterPurifier waterPurifier, float timeCooked)
- void OnWaterPurified(WaterPurifier waterPurifier, float timeCooked)

PR #303, opened 2021-04-29
Added OnIsOwned hook

senyaa1:
Added OnIsOwned hook to unlock _dance_01 (834887525)_ gesture.

Code Before:
```
public bool IsOwnedBy(BasePlayer player)
{
	return this.forceUnlock || (this.dlcItem != null && this.dlcItem.CanUse(player)) || (this.inventoryItem != null && player.blueprints.steamInventory.HasItem(this.inventoryItem.id));
}
```
Code After:
```
public bool IsOwnedBy(BasePlayer player)
{
	object returnvar = Interface.CallHook("CanUseGesture", this);
	if (returnvar != null)
	{
		return returnvar is bool && (bool)returnvar;
	}
	return this.forceUnlock || (this.dlcItem != null && this.dlcItem.CanUse(player)) || (this.inventoryItem != null && player.blueprints.steamInventory.HasItem(this.inventoryItem.id));
}
```

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -1,6 +1,6 @@
 {
   "Name": "Rust",
-  "TargetDirectory": "C:\\Users\\zloyk\\Desktop\\rustserv\\rustserv\\RustDedicated_Data\\Managed",
+  "TargetDirectory": "E:\\Servers\\Rust\\RustDedicated_Data\\Managed",
   "Manifests": [
     {
       "AssemblyName": "Assembly-CSharp.dll",
@@ -15625,7 +15625,7 @@
             "ArgumentString": "this",
             "HookTypeName": "Simple",
             "Name": "IsOwnedBy",
-            "HookName": "OnIsOwnedBy",
+            "HookName": "CanUseGesture",
             "AssemblyName": "Assembly-CSharp.dll",
             "TypeName": "GestureConfig",
             "Flagged": false,
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -1,6 +1,6 @@
 {
   "Name": "Rust",
-  "TargetDirectory": "E:\\Servers\\Rust\\RustDedicated_Data\\Managed",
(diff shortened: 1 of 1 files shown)


Calytic:
Couple things:
1. TargetDirectory shouldn't change, we won't be able to accept this as-is with the modified TargetDirectory
2. I think "OnIsOwnedBy" is too general a name. It isn't obvious _what_ is owned (I know it's Gestures) but it would be better if the hook name implicitly described what it was used for.

You tell me, but perhaps "CanUseGesture" would be better? If anyone has any suggestions here, please chime in.

senyaa1:
Changed hook name to "CanUseGesture" and changed TargetDirectory back to "E:\\Servers\\Rust\\RustDedicated_Data\\Managed"

0x89A:
In my opinion CanUseGesture doesn't really fit this hook since it is placed in the IsOwnedBy method which checks if the player owns the steam item or DLC. I would imagine a hook called CanUseGesture would follow the same pattern as other hooks like CanCraft etc.

Maybe something like PlayerOwnsGesture or something along those lines.

senyaa1:
Maybe PlayerOwnsGestureDLC would fit, or is it too long?

Calytic:
I would assume this method is called before a DLC gesture is used (to check if the player actually owns it). If that assumption is valid then CanUseGesture would be appropriate but you're right that it wouldn't apply to all gestures, only DLC gestures. So perhaps "CanUseDLCGesture"?

0x89A:
> I would assume this method is called before a DLC gesture is used (to check if the player actually owns it). If that assumption is valid then CanUseGesture would be appropriate but you're right that it wouldn't apply to all gestures, only DLC gestures. So perhaps "CanUseDLCGesture"?

Now that you say that, it looks like that it is called for every gesture

`private void Server_StartGesture(global::GestureConfig toPlay)
	{
		if (toPlay != null && toPlay.IsOwnedBy(this))
		{
		    ...
`

lukespragg:
Added, but did change to add the BasePlayer as an arg. Thanks!
- bool CanUseGesture(BasePlayer player, GestureConfig gesture)

senyaa1:
Thanks!

PR #310, opened 2021-05-19
Added OnCraneTakeDamageWhenNotInJunkyard hook

senyaa1:
### Added OnCraneTakeDamageWhenNotInJunkyard hook

Example:
```c#
bool? OnCraneTakeDamageWhenNotInJunkyard(BaseCrane crane)
{
    crane.GetDriver()?.IPlayer.Reply("OnCraneTakeDamageWhenNotInJunkyard works!");
    crane.lastDamagePos = crane.transform.position;
    return null;
}
```

**_But_**
` crane.lastDamagePos = crane.transform.position; ` is _**needed**_, otherwise it would crash

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -16023,6 +16023,30 @@
             "BaseHookName": null,
             "HookCategory": "Player"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 101,
+            "ReturnBehavior": 4,
+            "ArgumentBehavior": 1,
+            "ArgumentString": "",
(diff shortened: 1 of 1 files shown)


WheteThunger:
What exactly is the use case for this? Currently, I assume you can detect, scale or prevent damage to the crane by using the `OnEntityTakeDamage` hook. You could also increase the `maxDistanceFromOrigin` field of specific cranes to allow them to be driven farther, or set the `spawnOrigin` field to `Vector3.zero` to prevent the distance check.

senyaa1:
Thanks, didn't know that was possible.

PR #313, opened 2021-06-03
Add OnCupboardAuthorize hook for ItemModDeployable.OnDeploy

thibmo:
This is to also trigger the event when the player is authed as the TC is placed


lukespragg:
Merged, thanks!

PR #314, opened 2021-06-13
Add OnEntityDismounted [lite] hook. 

Slut:
You can do a "lite dismount" that doesn't check for a valid dismount position, but instead just dismounts the player presumably inside the entity. This is good for plugins that teleport people as they don't need a valid dismount location before teleporting them elsewhere. This means that OnEntityDismounted will never be called for people using a lite dismount.


lukespragg:
Merged, thanks!

PR #317, opened 2021-06-30
Ignore network limited players when updating acquaintances

WheteThunger:
This intends to address vanished players showing up as recently seen by other players.


lukespragg:
Merged, thanks!

PR #318, opened 2021-06-30
Add hook to prevent CCTV direction change

WheteThunger:
```cs
bool? OnCCTVDirectionChange(CCTV_RC camera, BasePlayer player)
```
- Called when a player attempts to set the direction of a CCTV camera
- Returning non-null cancels the default behavior


lukespragg:
Merged, thanks!

PR #322, opened 2021-07-09
Added new hook: CanDeliveryDroneAccessVendingMachine

Maelep:
The hook is called when a player open the marketplace so, if you revoke a Vending Machine, it will show "Unreachable" on the map.


lukespragg:
Merged with some changes, thanks!

SettLe1:
Hook "CanAccessVendingMachine" does not prohibit the purchase by drones.

PR #324, opened 2021-07-11
Add hook to ConsumptionAmount in AutoTurret

samrg472:
Exposes `ConsumptionAmount` as a field in `AutoTurret` to allow changing energy usage.

**IMPORTANT**: Requires PR OxideMod/Oxide.Patcher#150 to be merged which adds proper support for field injection.


lukespragg:
Merged, thanks!

PR #325, opened 2021-07-12
Add OnPlayerDismountFailed hook

WheteThunger:
```csharp
void OnPlayerDismountFailed(BasePlayer player, BaseMountable mountable)
```
Called when a player wants to dismount but there are no valid dismount positions.


lukespragg:
Merged, thanks!

PR #328, opened 2021-07-14
4 new hooks.

Maelep:
OnDecayTickDamage: Before decaying (damaging) entity.
OnDecayTickHeal: Before decaying (healing) entity.
OnWindUpdate: Before updating wind (= electricity power output)
OnWindUpdated: After updating wind (= electricity power output)

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -16212,8 +16212,8 @@
               }
             ],
             "HookTypeName": "Modify",
-            "Name": "FixAIBrainTryThinkNRE [patch]",
-            "HookName": "FixAIBrainTryThinkNRE [patch]",
+            "Name": "FixTryThinkNRE [patch]",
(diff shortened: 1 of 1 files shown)


WheteThunger:
I suggest making this two separate PRs, considering that decay and wind are unrelated. Smaller PRs generally get merged more easily.

Another tip would be to add the hooks to categories. You can do this via drag and drop in the Oxide Patcher UI.

Maelep:
> I suggest making this two separate PRs, considering that decay and wind are unrelated. Smaller PRs generally get merged more easily.
So I have to delete this one and then do it again or I can edit to separate in two PR?

> Another tip would be to add the hooks to categories. You can do this via drag and drop in the Oxide Patcher UI.
Ok, I'll do.

lukespragg:
Cleaned up and merged with some changes, thanks!

PR #331, opened 2021-07-17
Add new Hook CanMove [ElevatorLift]

Uhrendoktor:
Adds Hook CanMove to ElevatorLift to offer plugins the option to enable Vehicles to be lifted by an Elevator.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -16869,6 +16869,30 @@
             "BaseHookName": "OnWindUpdate",
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Simple",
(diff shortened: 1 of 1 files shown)


WheteThunger:
FYI: An alternative approach I have had success with is to disable the vehicle trigger with `lift.VehicleTrigger.SetActive(false)`.

Uhrendoktor:
> FYI: An alternative approach I have had success with is to disable the vehicle trigger with `lift.VehicleTrigger.SetActive(false)`.

Thanks. Nice approach. I wasn't aware of it.

lukespragg:
Thanks, merged!

PR #333, opened 2021-07-31
New hook OnVendingMachineRefill

jbrazio:
Solves an edge case leading to "Index Out of Bounds" on `NPCVendingMachine.Refill()` when the number of items inside a `NPCVendingMachine` are modified after the execution of `NPCVendingMachine.ServerInit()`.

`NPCVendingMachine.ServerInit()` starts a background task invoking `NPCVendingMachine.Refill()` every second. The first time it executes, `float[] refillTimes` gets initialized with the size of `this.vendingOrders.orders.Length` and it will never be re-evaluated for changes on the number of orders inside the vending machine during runtime.

This PR provides two changes:
1) Changes the visibility of `float[] refillTimes` to `public`
2) Adds a new `OnVendingMachineRefill(NPCVendingMachine machine)` hook

Usage example:
```cs
private object OnVendingMachineRefill(NPCVendingMachine machine)
{
	if (machine.vendingOrders.orders.Length != machine.refillTimes.Length)
		machine.refillTimes = new float[machine.vendingOrders.orders.Length];
	return null;
}
```

Or a more realistic alternative..
```cs
machine.vendingOrders.orders = myOrders.ToArray();
machine.refillTimes = new float[machine.vendingOrders.orders.Length];
machine.InstallFromVendingOrders();
```

I was unsure to name this hook `OnNPCVendingMachineRefill` but as another NPCVendingMachine hooks were already using the `[NPC]` suffix I went with it.

There is also a small ~~OCD~~ cosmetic change inside this PR which changes the display name of `CanAdministerVending [npc]` to uppercase `NPC`.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -7391,7 +7391,7 @@
             "ArgumentBehavior": 4,
             "ArgumentString": "a0, this",
             "HookTypeName": "Simple",
-            "Name": "CanAdministerVending [npc]",
+            "Name": "CanAdministerVending [NPC]",
             "HookName": "CanAdministerVending",
             "AssemblyName": "Assembly-CSharp.dll",
             "TypeName": "NPCVendingMachine",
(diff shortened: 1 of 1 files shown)


lukespragg:
Thanks, merged with minor changes!

PR #335, opened 2021-08-02
Added 6 metabolism damage hooks.

Maelep:


Maelep:
Can someone help me with this error? I don't understand. Is it because the method is `protected override` or because of `base.DoMetabolismDamage(ownerEntity, delta);` or something else?

PR #336, opened 2021-08-02
Add 7 metabolism damage hooks.

Maelep:


Maelep:
I don't understand from where the error comes.
Is it because of `protected override`?
Or because of `base.DoMetabolismDamage(ownerEntity, delta);`?
Or something else?

PR #337, opened 2021-08-05
New hook: CanItemBeRecycled

Vlad-00003:
In the last update devs added a way to prevent unrecyclable items from being put into the first 6 slots of recycler's inventory
This hook would allow plugins to change what can be placed there (useful for custom recyclables)


lukespragg:
Thanks, merged with minor changes!

PR #343, opened 2021-08-12
Add OnEntityPickedUp hook

WheteThunger:
```csharp
void OnEntityPickedUp(BaseCombatEntity entity, Item createdItem, BasePlayer player)
```
- Called after an entity has been picked up, before anything is done with the entity's inventory
- One use case is to clear the contents of the entity's inventory, to prevent those contents from going into the item's inventory or into the player inventory (e.g., to prevent a shotgun trap's ammo from being given to the player)

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -16947,6 +16947,33 @@
             "BaseHookName": null,
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 0,
+            "ReturnBehavior": 0,
+            "ArgumentBehavior": 3,
+            "ArgumentString": null,
+            "HookTypeName": "Simple",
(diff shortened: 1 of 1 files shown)


ChristianLW:
May I suggest changing the signature to the following?
```cs
void OnEntityPickup(Item createdItem, BasePlayer player, BaseCombatEntity entity)
```
It would fit better with the already existing `OnItemPickup` and `OnCollectiblePickup` hooks:
```cs
object OnItemPickup(Item item, BasePlayer player)
```
```cs
object OnCollectiblePickup(Item item, BasePlayer player, CollectibleEntity entity)
```

WheteThunger:
Those are named that way because they are pre hooks. This a post hook, so its naming should reflect that the pickup is already done or under way.

lukespragg:
Merged, thanks!

PR #344, opened 2021-08-12
Add OnPickupEntity

ChristianLW:
Currently there is a hook for disabling the ability to pickup a deployable (`CanPickupEntity`),
but there is no hook for after it has been picked up, and after the item has been created.
This PR adds that hook, giving the player, item, and entity as arguments, and with no return behaviour.
It runs after the item has been given, but before the deployed entity is destroyed.


ChristianLW:
I did not see the PR made 15 hours ago, oops.

PR #348, opened 2021-09-02
Added 5 fishing hooks.

ThePitereq:
Added 4 hooks to BaseFishingRod and 1 for WildlifeTrap.

object CanCastFishingRod(BasePlayer player, BaseFishingRod fishingrod, Item lure) - Called before player fishing rod cast.
void OnFishingRodCast(BasePlayer player, BaseFishingRod fishingrod, Item lure) - Called after player fishing rod cast.
object OnFishCaught(BasePlayer player, BaseFishingRod fishingrod, Item fish) - Called when player catch a fish.
void OnFishingInterrupted(BaseFishingRod fishingrod, BaseFishingRod.FailReason reason) - Called at fishing interrupted by fail.
object CanWildlifeTrap(WildlifeTrap trap, TrappableWildlife trapped) - Called when fish is caught to the fish trap.

Sorry for so much pull requests. I am still learning how to use Git :)

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -17049,6 +17049,134 @@
             "BaseHookName": null,
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 10,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
+            "ArgumentString": "l1, this, l2",
+            "HookTypeName": "Simple",
+            "Name": "CanCastFishingRod",
+            "HookName": "CanCastFishingRod",
+            "AssemblyName": "Assembly-CSharp.dll",
+            "TypeName": "BaseFishingRod",
+            "Flagged": false,
+            "Signature": {
+              "Exposure": 0,
+              "Name": "Server_RequestCast",
+              "ReturnType": "System.Void",
+              "Parameters": [
+                "BaseEntity/RPCMessage"
+              ]
+            },
+            "MSILHash": "XLRgAS3RdQN/Nd3iaBndgpPPtM5u02eIr+b8AUwGBTg=",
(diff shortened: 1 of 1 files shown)


lukespragg:
Haven't forgotten about you! There are a couple issues we're discussing regarding this, but there'll likely need to be a few changes with return handling and naming, as well as some additional post hooks where they make sense.

ThePitereq:
Hey. I set the names according to the docs. If something is wrong, then sorry. All hooks worked for me in the way as I wanted. It's not mentioned there, but OnFishCaught can return fish that is caught. I've not made any more additional hooks, because i though that is enough, and I didn't have any idea for implementing more :)

lukespragg:
This is what we ended up with, let me know if anything doesn't make sense!

- bool CanCastFishingRod(BasePlayer player, BaseFishingRod fishingRod)
- bool CanCatchFish(BasePlayer player, BaseFishingRod fishingRod, Item fish)
- object OnWildlifeTrap(WildlifeTrap trap, TrappableWildlife trapped)
- void OnFishCaught(Item fish, BaseFishingRod fishingRod, BasePlayer player)
- void OnFishingRodCast(BaseFishingRod fishingRod, BasePlayer player)
- void OnFishingStopped(BaseFishingRod fishingRod, BaseFishingRod.FailReason reason)

ThePitereq:
I see that you've removed lure variable. Maybe I am blind but i didn't see any available way to get the item that has been used as lure (I am not talking about fishing rod inventory) on fishing rod cast. Besides, everything seems to be correct. Thank you! :)

lukespragg:
Likely just missed it! I'll take a look.

ThePitereq:
I think it is possible, to check inventory at CanCastFishingRod and then save value and go to OnFishCaught, but i think adding it will be better. ^^

PR #349, opened 2021-09-14
ServerMgr: Expose field and add restart-related hooks

2chevskii:
## Exposed field:
`ServerMgr.restartCoroutine` - stores coroutine used for server restarts
### ServerMgr::RestartServer

- `OnServerRestart(string notice, int delaySeconds): Exit if non-null` Gets called right before restart coroutine is initialized, allows to cancel default behaviour.
- `OnServerRestartInterrupt(): Exit if non-null` Gets called after the check if restart coroutine != null and before server "Restart interrupted!" message, allows to cancel both interruption and following re-initialization of restart process.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -17049,6 +17049,60 @@
             "BaseHookName": null,
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Simple",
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged, thanks!

PR #351, opened 2021-09-21
RustExtension: Add GetLatestVersion method to retrieve last Oxide.Rust build version

2chevskii:
## Purpose
Allow plugins to painlessly get last official build version of Oxide.Rust in ready-to-go format (VersionNumber). 
Useful for cases where plugin developer wants to implement some kind of update notifier.
### `public GetLatestVersion(Action<VersionNumber, Exception>): void`
Method to get last extension version
### `private async GetLatestExtensionVersion(): Task<VersionNumber>`
Private method which implements logic to get information from GitHub API
### `private ParseVersionNumber(string): VersionNumber`
Private method for parsing API response into VersionNumber
*Should be considered a temporary solution, definitely would be better to add parsing functional into VersionNumber itself*


lukespragg:
Merged, thanks!

PR #352, opened 2021-10-01
Add new hooks

rostov114:
**CanItemPurchased** - executed before the Item is passed to the player or drone.
**OnPlayerDrink** - executed when the player tries to drink
**OnApplicationCommand** - executed when the player tries to send a chat command from Rust+
**OnUserApplicationCommand** - same as **OnApplicationCommand**, covalence.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -17402,6 +17402,103 @@
             "BaseHookName": null,
             "HookCategory": "NPC"
           }
+        },
+        {
+          "Type": "Modify",
+          "Hook": {
+            "InjectionIndex": 241,
+            "RemoveCount": 0,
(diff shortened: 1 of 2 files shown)


rostov114:
@lukespragg please check the **CanItemPurchased** hook more precisely, there is a suspicion that I lost when playing with the position of the IL code.

lukespragg:
You should be able to use the same hook name for OnApplicationCommand with both calls. I'll take a look at the IL as well as the others once I get a chance.

lukespragg:
Also, CanItemPurchased should probably be named CanPurchaseItem or OnItemPurchase. OnPlayerDrink seems fine but should probably have the player argument first.

PR #354, opened 2021-10-08
Add OnOpenVendingShop hook to NPCTalking

WheteThunger:
This adds the existing `OnOpenVendingShop` hook to the `NPCTalking` class so that it is also called when a player opens an invisible vending machine from an NPC conversation.

```cs
void OnOpenVendingShop(VendingMachine machine, BasePlayer player)
```

The ideal naming would be `OnVendingShopOpen`, to be consistent with the new standards, but this is an existing hook already used in another class, so it would not make sense to change the name for this usage.


lukespragg:
Merged, thanks!

PR #359, opened 2021-10-29
Add OnWaterCollect [WaterPump] hook

Substrata1:
```csharp
object OnWaterCollect(WaterPump waterPump, ItemDefinition water)
```
Called before water is added to a water pump


lukespragg:
Merged, thanks!

PR #361, opened 2021-10-31
Add CanUpdateSign hook for CarvablePumpkin

WheteThunger:
```cs
bool? CanUpdateSign(BasePlayer player, CarvablePumpkin pumpkin)
```

This hook uses the exact same implementation as the hook of the same name in the `Signage` and `PhotoFrame` classes.

Tested in-game and confirmed that carvable pumpkins can be updated when returning `null`, and not updated when returning `false`.


lukespragg:
Merged, thanks!

PR #362, opened 2021-11-04
Expose Fields / Methods on SamSite

dassjosh:
Expose Fields / Methods on SamSite
Method SetTarget
Method ClearTarget
Field mostRecentTargetType
Field currentTarget


lukespragg:
Thanks, merged manually to avoid the target change. ;)

PR #366, opened 2021-11-16
Added OnLootSpawn hook to ItemModUnwrap 

k1lly0u:
Added OnLootSpawn hook to ItemModUnwrap prior to spawning loot and consuming item


lukespragg:
Merged as object OnItemUnwrap (Item item, BasePlayer player, ItemModUnwrap wrappedItem), thanks!

PR #368, opened 2021-11-17
Expose various methods/fields as public

mrcameron999:
- Expose various fields and a class to be public for building stability

- Made StabilityEntity.Support class public, This class is a nested class

- Exposed BaseEntityl.links field to be public
`System.Collections.Generic.List<EntityLink> links = new System.Collections.Generic.List<EntityLink>();`

- Exposed StabilityEntity.supports field to be public
 `List<StabilityEntity.Support> supports;`

All these fields/classes are only used for building and are used when working out stability for building blocks.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -40524,6 +40524,63 @@
             "Parameters": []
           },
           "MSILHash": ""
+        },
+        {
+          "Name": "StabilityEntity::supports",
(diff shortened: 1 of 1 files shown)


lukespragg:
There was some concern expressed that the List<StabilityEntity.Support> field might break something since there are null checks for that field in the parent class.

lukespragg:
Merged all except for the one mentioned, which we'll have to look into yet.

PR #370, opened 2021-12-01
Add Xmas event hooks

Substrata1:
```csharp
object OnDistributeXmasLoot(XMasRefill xmas)
```
- Called when the Xmas event runs, before loot is distributed

```csharp
object OnFillXmasStocking(Stocking stocking)
```
- Called when the Xmas event runs, before a stocking is filled with loot

```csharp
object OnGiveXmasGifts(XMasRefill xmas, BasePlayer player)
```
- Called when the Xmas event runs, before gifts are spawned for players

Expose SantaSleigh.dropPosition

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -17977,6 +17977,80 @@
             "BaseHookName": null,
             "HookCategory": "Item"
           }
+        },
+        {
+          "Type": "Simple",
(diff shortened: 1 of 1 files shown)


lukespragg:
Manually merged with changes to names and return for one, thanks!

PR #372, opened 2021-12-03
Reinstate OnCorpsePopulate hook for ScarecrowNPC

k1lly0u:


lukespragg:
Added along with the others as discussed, thanks!

PR #378, opened 2021-12-16
BaseAIBrain Senses setter to public

jalamaack:
Set BaseAIBrain Senses  setter to public


lukespragg:
Merged in BaseAIBrain`1::Senses setter as public, thanks!

PR #379, opened 2021-12-16
Expose EntityFuelSystem.GetFuelContainer() as public

WheteThunger:
This method is already `public` in today's Rust, but it's `private` in tomorrow's Rust, so this modifier should ensure it stays public. This should be merged prior to the next Oxide update to avoid breaking many plugins.


lukespragg:
Thanks, merged with staging changes!

PR #386, opened 2022-01-04
Add OnCreateProjectileEffectClientside hook

senyaa1:
```csharp
object OnCreateProjectileEffectClientside(BaseProjectile projectile, string prefabName, Connection sourceConnection)
{
    Puts("OnCreateProjectileEffectClientside works!");
    return null;
}
```

- Allows hiding projectile effects


lukespragg:
Merged as object OnClientProjectileEffectCreate(Connection connection, BaseProjectile projectile, string prefabName)

Thanks!

PR #394, opened 2022-02-09
Add hooks and expose Sprinkler::currentFuelType

ArtiIOMI:
void OnWaterBallDoSplash(Vector3 position, float radius, ItemDefinition liquidDef, int amount)
_- Called when water is poured from the vessel or we shoot water from a water gun._

void OnSprinklerSplas(Sprinkler sprinkler)
_- Call when the sprinkler starts watering._


lukespragg:
Going to have to merge this manually, but in the future, please make sure to use a branch just for your PR (not develop), and rebase rather than committing new commits manually.

PR #397, opened 2022-02-22
Add SprayCan hooks

WheteThunger:
```cs
object OnEntityReskin(BaseEntity entity, ItemSkinDirectory.Skin skin, BasePlayer player)
```
- Called when a player selects a skin with the spray can
- Returning non-null cancels the default behavior

```cs
void OnEntityReskinned(BaseEntity entity, ItemSkinDirectory.Skin skin, BasePlayer player)
```
- Called after an entity has been reskinned
- No return behavior


lukespragg:
Merged, thanks!

PR #398, opened 2022-02-28
Improving user language support.

Maelep:
By default Oxide only allow to change your language to an existing language folder meaning that the user language will never be supported if the user sets his language before a plug-in supporting his language is installed (except if he try again to sets his language).

My change will provide the user language as soon as the user sets his language and a plug-in supports it. No error will be caused since if the language is not supported then Oxide returns the English version by default.

I added a basic condition to avoid weird language.

### DIFF
--- a/src/RustCommands.cs
+++ b/src/RustCommands.cs
@@ -232,8 +232,7 @@ namespace Oxide.Game.Rust
             else
             {
                 // TODO: Check if language exists before setting, warn if not
-                string[] languages = lang.GetLanguages();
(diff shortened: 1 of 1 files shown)


lukespragg:
It may be better to check if the provided input is a valid culture rather than using regex.

Maelep:
> It may be better to check if the provided input is a valid culture rather than using regex.

Thank you, you made me discovered that. I'll fix dumb the error (too tired...) and make the improvement.

PR #400, opened 2022-03-01
Queue Hooks

TravisDevLife:


lukespragg:
Merged, thanks! Had to do some tweaks with the return types, removed the parameter specification (there's an option for that, JustParams), and added a description. If you'd like to take a look, here's the commit: https://github.com/OxideMod/Oxide.Rust/commit/a780da47c4d98fac78ab91ec0ace06c070bb4d02

Also, made the OnConnectionDequeue returnable as it is a pre hook the way it is positioned and named right now.

PR #401, opened 2022-03-25
Add OnEngineStartFinished hook

WheteThunger:
```cs
void OnEngineStartFinished(BaseVehicle vehicle)
```
Called after a vehicle's engine has actually finished starting.


lukespragg:
Merged, thanks!

PR #403, opened 2022-04-03
Add OnNetworkSubscriptionsGather hook

WheteThunger:
```cs
object OnNetworkSubscriptionsGather(NetworkVisibilityGrid grid, Network.Visibility.Group group, List<Network.Visibility.Group> groups, int radius)
```
- Called when determining which network groups to subscribe a player to in a given radius
- Returning non-null will cancel the default behavior


lukespragg:
Merged, thanks again!

PR #405, opened 2022-04-06
Added PaintedItemStorageEntity Hook

dassjosh:
Added void OnItemPainted(PaintedItemStorageEntity entity, BasePlayer player, Item item, byte[] image)
Exposed _currentImageCrc


lukespragg:
Merged manually, thanks!

PR #407, opened 2022-04-17
Add OnRecycleAmount hook

birthdates:
```csharp
object OnRecycleAmount(Item item, int defaultAmount, Recycler recycler)
```
Change the amount of an item that is recycled per tick (i.e by default, you can only recycle one rifle body every tick. However, if you return 5, it will recycle 5)


lukespragg:
Merged, thanks!

PR #412, opened 2022-05-30
Add hook for SamSite ToggleDefenderMode

ThePaperPlate:
- bool? CanToggleDefenderMode(SamSite samSite, BasePlayer player, bool enable)


lukespragg:
Thanks! Your naming didn't quite match out normal hook pattern, so this is what we ended up with:

- object OnSamSiteModeToggle(SamSite samSite, BasePlayer player, bool enabled)

PR #414, opened 2022-06-06
Added OnLockerSwap hook.

ThePitereq:
Added an OnLockerSwap hook to the locker executed when pressing the "Swap" button.

`OnLockerSwap(Locker locker, BasePlayer player, int buttonIndex)`


lukespragg:
Merged, but bumped up a little bit to be above the IsEquipping check and moved the player arg to last. Thanks!

PR #418, opened 2022-06-13
Add OnSprayRemove hook and splashThreshold field

WheteThunger:
- Added `object OnSprayRemove(SprayCanSpray spray, BasePlayer player)` hook, which can be used to prevent a player from washing away a spray while holding a water item
- Added the `splashThreshold` field to the `SprayCanSpray` class, which determines how much water in ml is required to wash away a spray (e.g., when deselecting a bucket of water)
- Exposed various fields/methods for spray entities

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -19085,6 +19085,69 @@
             "BaseHookName": null,
             "HookCategory": "Weapon"
           }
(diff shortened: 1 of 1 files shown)


lukespragg:
Merged, thanks!

PR #424, opened 2022-07-12
Add OnPlayerReport

SolidSnake-72:
Add OnPlayerReport pre-hook to allow cancelling

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -45497,6 +45497,42 @@
           "Flagged": false
         }
       ]
+    },
+    {
+      "AssemblyName": "EasyAntiCheat.Server.dll",
+      "Hooks": [
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 0,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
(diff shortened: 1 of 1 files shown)


lukespragg:
What would be the purpose of this hook? Like, when would this be necessary?

SolidSnake-72:
To either allow or block the sending of player reports, based on player. Currently there is no way to do this for players who abuse the report system. I also see a lot of players sending reports based on server-specific rule violations, not realizing this goes to Facepunch/EAC for game ban consideration.

lukespragg:
So one thought was to hook the 'cheatreport' ConVar instead of the other method to avoid the extra DLL being patched, which would probably be fine given that's the only place I see that calls that method right now. Thoughts?

soulfly006:
They want to intercept and cancel the report…which is not what facepunch wants

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

SolidSnake-72:
I did try to hook the ConVar initially, but unfortunately that only works for cheat reports, not abuse or spam.

lukespragg:
Based on what others have suggested, it sounds like this wouldn't have any benefit being as the client sends directly to Facepunch.

PR #427, opened 2022-08-11
Add OnRCUpdateIdentifier hook

senyaa1:
```csharp
object OnRCUpdateIdentifier(PoweredRemoteControlEntity entity, string newID, bool clientSend)
```

- Allows cancelling RC identifier update by returning non-null value


lukespragg:
Hey, thanks for the PR! Made a few tweaks to follow our naming and such, which would result in this:

- object OnRemoteIdentifierUpdate(PoweredRemoteControlEntity entity, string newId)

The last arg was not used, so it was dropped, and the return type was changed (still the same result).

PR #430, opened 2022-09-08
Adds a new BaseOven OnGetTemperature hook.

jbrazio:
Adds a new BaseOven OnGetTemperature hook:
```cs
	object returnvar = Interface.CallHook("OnGetTemperature", this, slot);
	if (returnvar is float)
	{
		return (float)returnvar;
	}
```


MrBlue:
Merged with some changes, OnGetTemperature is now OnOvenTemperature.
Thanks!

PR #431, opened 2022-09-18
 Added Hook void OnBuildingMerge(BuildingManager.Building to, BuildingManager.Building from)

dassjosh:
- Added Hook void OnBuildingMerge(BuildingManager.Building to, BuildingManager.Building from)


MrBlue:
Merged, thanks!

PR #432, opened 2022-09-20
AI and target manipulation additions

read-0nly:
Adds the following section
- AI - maybe it's best for it to be in NPC, but I felt like it might lead to confusion with other NPC things. Happy to change this and refactor on my end if you guys don't agree that a separation between stuff specifically targeting the AI brain and stuff targeting the AI body in space is useful. Could use suggestions for hook names.

Adds the following hooks
- OnAICaresAbout - Allows modifying the function by which entities are filtered to be tracked by AI. Combined with the hook in SetKnown, allows setting unusual targets for native AI
- OnAIInitialize - triggers after the AI has loaded it's AI Design but before the brain starts ticking - allows replacing states safely
- OnNpcGetBestRoamPoint - lets you give back an arbitrary AIMovePoint during Scientist Roam, so they can break out of AIZs
- OnTrapInterest - allows injection of unusual trap targets
- OnTurretShouldTarget - Allows injection of unusual turret targets
- OnTurretCheckHostile - Allows manipulation of turret hostility detection (useful in conjunction with OnTurretShouldTarget to make them react to NPCs)
- OnHelicopterUpdateTargets - Allows injection of helicopter targets (like shooting at NPCs)

Adds the following modifiers
- BaseNavigator::navMeshQueryFilter to public - Useful for ill-advised navmesh manipulations (I haven't found a way around needing this even if it's use is niche)
- "BaseNavigator::defaultAreaMask to public - literally the same reason.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -18446,7 +18446,7 @@
                 "Epic.OnlineServices.AntiCheatCommon.OnClientActionRequiredCallbackInfo&"
               ]
             },
-            "MSILHash": "bBFuplhWlqfMN9pHvi4XIEI3tSPghjVi7Ds4s4PAiEw=",
+            "MSILHash": "ERNOvg1y3Bk+7GT2o48RKLV30BTruiYXJHE+a35AdwQ=",
             "HookCategory": "Player"
           }
         },
@@ -18471,7 +18471,7 @@
                 "Epic.OnlineServices.AntiCheatCommon.OnClientActionRequiredCallbackInfo&"
               ]
             },
-            "MSILHash": "bBFuplhWlqfMN9pHvi4XIEI3tSPghjVi7Ds4s4PAiEw=",
+            "MSILHash": "ERNOvg1y3Bk+7GT2o48RKLV30BTruiYXJHE+a35AdwQ=",
             "BaseHookName": "OnPlayerBanned [EAC]",
             "HookCategory": "Player"
           }


read-0nly:
A lot of this is ultimately about getting NPCs to go where they don't belong and shoot at things they shouldn't using the vanilla brain loop instead of building a secondary solution. Already used it to create a faction system where bandit guards, scientists, and dwellers have a custom event/state map and custom brain states that allow them to roam the map autonomously and engage each other over territory, all using the vanilla brain loop.

If I can get at least most of these approved it'll be followed up with plugins to umod for devs to better integrate with this themselves, for a bit more of an organic feel on custom behaviours.

read-0nly:
Some examples of what the hooks let you do, with my psilocybin statemap. This is all very rough - I literally finished the first draft of Omninasty and VehicleHandler yesterday - I'm in the middle of revamping all of this so it's actually modular and not a single 3k line file. They'll roam, fight, and if you get in a vehicle and turn it on, if they're roaming, they'll hop in with you. If they're in chase state they'll dismount. There are probably bugs. Grab an NPC buddy with a minicopter then run the minicopter into a phone pole - they get yeeted across the map

[plugins.zip](https://github.com/OxideMod/Oxide.Rust/files/9617674/plugins.zip)
[Psilocybin.json.txt](https://github.com/OxideMod/Oxide.Rust/files/9617692/Psilocybin.json.txt)

read-0nly:
Darn, trying to fix the merge issue closed this. I'll wait till after wipe, then re-submit

PR #433, opened 2022-09-22
Add OnImpactEffect and OnEyePosValidate hooks

senyaa1:
```csharp
bool? OnEyePosValidate(AttackEntity entity, BasePlayer player, Vector3 eyePos)
```

- Returning true bypasses eye position checks

```csharp
object OnImpactEffect(HitInfo info)
```

- Returning non-null value disables impact effects

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -16210,6 +16210,56 @@
           }
         },
         {
+          "Type": "Simple",
+          "Hook": {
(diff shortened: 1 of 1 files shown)


MrBlue:
Merged with some changes, OnImpactEffect is now OnImpactEffectCreate.

Thanks!

PR #434, opened 2022-09-26
Add Patches to CombatLog.Log

dassjosh:
- Add Patches to CombatLog.Log due to recent changes not correctly checking unity game object for null causing RPC Error in OnProjectileAttack

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -18544,6 +18544,81 @@
             "BaseHookName": "OnOvenStart",
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Modify",
+          "Hook": {
+            "InjectionIndex": 109,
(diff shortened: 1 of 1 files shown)


WheteThunger:
When are people encountering the `RPC error in OnProjectileAttack`? In conjunction with plugins that are deleting the held entity while a projectile is in flight, or in vanilla cases as well?

dassjosh:
This has been confirmed to be a modded issue. Instead of trying to update all plugins that have this issue figured it would be easier to just patch in the correct checks.

shaitobu:
> 

or if player dies/disconnects while projectile is in flight

MrBlue:
This will be fixed by Facepunch next update! (Thanks @Jake-Rich!)

PR #436, opened 2022-10-23
Add OnPortalUse hook

senyaa1:
```csharp
object OnPortalUse(BasePortal portal, BasePlayer player)
```

- Returning non-null value cancels player teleport


MrBlue:
Merged, thanks!

PR #437, opened 2022-10-24
Add hook handle unknown chat command

rostov114:
To create dynamic unregistered chat commands.

### DIFF
--- a/src/RustHooks.cs
+++ b/src/RustHooks.cs
@@ -350,6 +350,11 @@ namespace Oxide.Game.Rust
                 // Is it a valid chat command?
                 if (!Covalence.CommandSystem.HandleChatMessage(basePlayer.IPlayer, str) && !cmdlib.HandleChatCommand(basePlayer, cmd, args))
                 {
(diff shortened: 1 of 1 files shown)


MrBlue:
What's the exact use-case of this hook?

rostov114:
1. Player create string data (example: `/create test`), save in data file 
2. Player(s) using in chat command: `/test`

MrBlue:
You can already dynamically register commands and there are already plugins like https://umod.org/plugins/custom-chat-commands that do this, I don't see a reason for this hook to exist...

austinv900:
@MrBlue someone in discord also wanted the ability to log any command even nonexistent ones. Oxide doesn't really allow that right now.

MrBlue:
> 

The OnPlayerCommand hook gets called before any command handling is done. So you should be able to use that to run any command ran by players, even nonexistent ones.

rostov114:
> The OnPlayerCommand hook gets called before any command handling is done. So you should be able to use that to run any command ran by players, even nonexistent ones.

Sometimes it is necessary to intercept processing in this place. 
**OnPlayerCommand** - low performance, because it registers for any command call.

PR #438, opened 2022-10-24
Added OnFeedbackReported hook.

ThePitereq:
Added new hook on feedback report.

`void OnFeedbackReported(BasePlayer player, string subject, string message, ReportType type)`


MrBlue:
Merged, thanks!

PR #439, opened 2022-11-03
Added Recreate property for new UI functionality

Kulltero:
https://github.com/Facepunch/Rust.Community/pull/40 is set to be added on force wipe, adding a new property to the panel if the key gets added the existing element gets destroyed (if it exists), this basically just saves a DestroyUI call.

i added the NullValueHandling Attribute since the pull request uses ContainsKey instead of GetBoolean, feel free to alter this if there is a better way of doing this

### DIFF
--- a/src/RustCui.cs
+++ b/src/RustCui.cs
@@ -193,6 +193,9 @@ namespace Oxide.Game.Rust.Cui
         [JsonProperty("parent")]
         public string Parent { get; set; }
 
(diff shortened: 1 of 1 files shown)


Kulltero:
you're completely correct, making the change now (somehow completely missed this so thank you for catching this, i'll chalk it up to having no sleep)

MrBlue:
Has this been merged into the game yet?

Kulltero:
yes it has ;) it was merged on the december forcewipe

MrBlue:
This was merged last update, thanks!

PR #443, opened 2022-12-27
OnMessagePlayer hook for Oxide.Game.Rust.Libraries.Player library

Vlad-00003:
This hook already exists in the game itself, on BasePlayer.ChatMessage method So I think it would make sense to call the same hook in the method, that does pretty much the same thing


MrBlue:
Merged, thanks!

PR #448, opened 2023-02-09
Added OnInvalidPositionCheck hook

senyaa1:
```csharp
object OnInvalidPositionCheck(BaseEntity entity)
{
    Puts("OnInvalidPositionCheck works!");
    return null;
}
```
- This hook allows disabling entity position check.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -589,7 +589,7 @@
             "HookName": "IOnUserApprove",
             "AssemblyName": "Assembly-CSharp.dll",
             "TypeName": "ConnectionAuth",
-            "Flagged": false,
+            "Flagged": true,
             "Signature": {
               "Exposure": 2,
               "Name": "OnNewConnection",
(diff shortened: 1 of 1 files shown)


MrBlue:
What is the use case of this hook exactly?

senyaa1:
> What is the use case of this hook exactly?

I was writing a plugin that required scaled entities. So the best _(and probably the only way)_ I came up with was parenting entities to a scaled SphereEntity. I've put it -1000 meters below the entity, so it would be invisible _(without using performance-heavy hooks like in [EntityScaleManager](https://umod.org/plugins/entity-scale-manager))_. But it despawns because sometimes it happens to be invalid position. So I've created a hook that allows disabling this check. (it also might be useful to detect items which fell through the map).

MrBlue:
You should be able to set -499 to avoid any issues, I don't see a reason to add this hook for this purpose.

PR #449, opened 2023-02-17
Updated Covalence server.port to use query port.

dassjosh:


dassjosh:
Rust changed the meaning of the ports. Query port took over what port used to be. Just updating it since there are plugins that use server.port which no longer works because of the change.

austinv900:
> Rust changed the meaning of the ports. Query port took over what port used to be. Just updating it since there are plugins that use server.port which no longer works because of the change.

I'd have to play with it when I get home. But from what I can see, everything on my server everything is working as intended.

These are my ports
- game 28015
- rcon 28015
- rust+ 28016
- query 28016

They are all working as expected.

0x89A:
Thanks for this contribution. However, I feel that it would be best to introduce a separate property for the query port. The reason for this is to avoid confusion with naming, and in the case that a developer might want this to be the intended behavior.

skythirix93:
The question I ask myself is that since Thursday I keep looking at how to apply that configuration on my dedicated and I have not had the luck to get it.


Currently I have managed to put the server online but the Steam Connect does not work and I do not know if it is because of rust or a misconfiguration.

I have contacted rust support and they have not wanted to give me a solution.

I have the following configuration

Game Port 28016 udp
Query Port 28015 udp
Rust ++ 28083 tcp
Rcon 28017 tcp

I leave my discord if you want to send me a private message or something. 
roger93#9737

MONaH-Rasta:
It would be interesting to know what ConVar.Server.port can now be used for

austinv900:
> It would be interesting to know what ConVar.Server.port can now be used for

It's still the port used to connect

MrBlue:
If anyone wants to make a PR to add an additional field to the server object called something like queryPort or similar feel free to do so, closing this as we don't want to change the existing implementation.

PR #450, opened 2023-02-24
- Added OnPrintToChat(BasePlayer player, string format, object[] args)

mbedda:
Added OnPrintToChat(string format, object[] args) & OnPrintToChat(BasePlayer player, string format, object[] args)

### DIFF
--- a/src/RustPlugin.cs
+++ b/src/RustPlugin.cs
@@ -148,7 +148,7 @@ namespace Oxide.Plugins
         /// <param name="args"></param>
         protected void PrintToChat(BasePlayer player, string format, params object[] args)
         {
-            if (player?.net != null)
+            if (player?.net != null && Interface.CallHook("OnPrintToChat", player, format, args) == null)
             {
                 player.SendConsoleCommand("chat.add", 2, 0, args.Length > 0 ? string.Format(format, args) : format);
             }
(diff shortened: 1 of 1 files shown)


MrBlue:
Hey, what's the reason for those hooks being added? Can't you use the `OnSendCommand` and `OnBroadcastCommand` hooks for this already?

mbedda:
`OnSendCommand` and `OnBroadcastCommand` do not intercept messages, which if they did they would cover all chat messages as most call `BasePlayer.SendConsoleCommand`. 

An example of this being used is this https://codefling.com/plugins/chat-to-notifications, were I convert some chat messages to notification to clear plugins chat spam. Here i had to use Harmony to intercept at `SendConsoleCommand` and `PrintToChat`, doing that for `ConsoleNetwork` functions was not great performance wise as it handles all commands (not only "chat.add" & "chat.add2" "say"..)

MrBlue:
Both `PrintToChat` methods call either `OnSendCommand` or `OnBroadcastCommand` internally so it would intercept any message being sent using the `PrintToChat` message?

mbedda:
Intercept meaning they do not override default behavior

MrBlue:
Okay so what you're saying is we should make the `OnSendCommand` and `OnBroadcastCommand` have a return behavior?

mbedda:
yes exactly

MrBlue:
Both OnSendCommand hooks and the OnBroadcastCommand hooks will now cancel when returning a non null value

PR #452, opened 2023-03-05
Add CanExplosiveStick hook

WheteThunger:
### Purpose
To bring back C4 drones.

### Usage
```cs
object CanExplosiveStick(TimedExplosive explosive, Drone drone) => true;
```

### Method being hooked
```cs
public virtual bool CanStickTo(BaseEntity entity)
{
    object obj = Interface.CallHook("CanExplosiveStick", this, entity);
    if (obj is bool)
    {
        return (bool)obj;
    }
    if (entity.TryGetComponent<DecorDeployable>(out var _))
    {
        return false;
    }
    if (entity is Drone)
    {
        return false;
    }
    return true;
}
```


MrBlue:
Merged, thanks

PR #453, opened 2023-03-27
Add OnInventoryAmmoItemFind hook

WheteThunger:
```cs
Item OnInventoryAmmoItemFind(PlayerInventory inventory, ItemDefinition itemDefinition)
```
- Called when reloading a `Chainsaw` or `FlameThrower`
- Returning an `Item` makes the weapon reload with that item instead of looking in the player inventory

This is similar to the existing hook `object OnInventoryAmmoFind(PlayerInventory inventory, List<Item> collect, AmmoTypes ammoType)`, but returns a single item.


MrBlue:
Merged, thanks!

PR #456, opened 2023-04-13
OnWireConnect hook fix & removal of the patch that nulls out linePoints

Nikedemos:
So as it turns out, after April 2023 Rust update, linePoints are assigned to the output IOentity AFTER calling the OnWireConnect hook. Mr. Blue suggested adding an argument that passes the list of linePoints to the hook call so the subscribed method can process it for whatever reason. Before April 2023, the method could just take outputEntity[outputSlot].linePoints, so this quick and dirty solution should help. It won't break anything if your hook subscription doesn't use it, since it's the last argument). Mr. Blue also agreed with me that removing the patch named "OnWireConnect [patch]" which added nulling out of the linePoints if the hook returned non-null - now the linePoints are assigned to the output entity afterwards, so it's not necessary - since nothing would get assigned in case of returning non.null.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -7469,7 +7469,7 @@
             "InjectionIndex": 167,
             "ReturnBehavior": 1,
             "ArgumentBehavior": 4,
-            "ArgumentString": "l0, l8, l4, l10, l6",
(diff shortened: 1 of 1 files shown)


MrBlue:
Merged, thanks

PR #461, opened 2023-05-08
Add altMove parameter to CanMoveItem

Substrata1:
```csharp
object CanMoveItem(Item item, PlayerInventory playerLoot, ItemContainerId targetContainer, int targetSlot, int amount, bool altMove)
```
altMove is true when using alt+right-click to directly equip clothing items. Required now for GetIdealContainer(), GiveItem() and GetIdealPickupContainer()


Nikedemos:
Sounds useful indeed, thumbs up

MrBlue:
Merged, thanks!

PR #463, opened 2023-06-02
Exposing BuildingBlock::playerCustomColourToApply for modding

collectVood:
This field is used to set the custom color for the new container skin and likely future skins.


MrBlue:
Merged, thanks!

PR #469, opened 2023-07-31
Added ItemCrafter argument to hooks

dassjosh:
Added ItemCrafter argument to OnItemCraftFinished and OnItemCraftCancelled hooks


MrBlue:
Merged, thanks!

PR #470, opened 2023-08-02
Add OnBedMade hook

WheteThunger:
```cs
void OnBedMade(SleepingBag sleepingBag, BasePlayer player)
{
    Puts("OnBedMade works");
}
```
Intended for detecting when a player "makes" a camper bed, to adjust respawn timers.


MrBlue:
Merged, thanks!

PR #471, opened 2023-08-22
Added Hook OnEntityLoaded

TurEduard:
Hook called after a network object is loaded from a save file. Has BaseNetworkable.LoadInfo as argument. Needed to fix the createFrame variable, including.


MrBlue:
Merged, thanks!

PR #475, opened 2023-09-22
Added OnFastTrackTask Hook

shooter46:
```
object OnFastTrackTask(ItemCrafter itemCrafter, ItemCraftTask task)
{
	Puts("OnFastTrackTask works!");
	return null;
}
```


MrBlue:
Merged, renamed to CanFastTrackCraftTask. Thanks!

PR #479, opened 2023-10-19
Add OnPatrolHelicopterTakeDamage hook

Substrata1:
```csharp
object OnPatrolHelicopterTakeDamage(PatrolHelicopter heli, HitInfo hitinfo)
```
Called when the Patrol Helicopter takes damage

This functionality cannot be replicated with OnEntityTakeDamage due to the way Patrol Helicopter deaths occur. That hook calls too late, and the helicopter is considered dead (heli.myAI.IsDead) on the final hit, even though it is not. This means you cannot accurately track total damage received during its pre-death state, or check when the final hit actually occurs, in order to get the killer.


MrBlue:
Merged, thanks!

PR #481, opened 2023-11-15
OnCoalingTowerBeginUnload and OnCoalingTowerUnload Hooks

Vlad-00003:
I'm not sure if what I did was the best option, because OnCoalingTowerBeginUnload is called before an actual attempt to unload happened.
OnCoalingTowerUnload can't exit due to the fact that it's in the middle of a iteration of polled list. There might be a better way but I wasn't able to fine one.

I need to know who began unloading and what resources being unloaded at the moment


MrBlue:
Merged with changes, thanks!

```cs
void OnCoalingTowerStart(CoalingTower tower, BasePlayer player)
void OnCoalingTowerGather(CoalingTower tower, Item item)
```

PR #483, opened 2023-11-24
Added hook "OnNoteUpdate" to keep track of changes or prevent them

21mac:
Needed the possibility to prevent changes to notes for a plugin I am working on.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -19791,7 +19791,31 @@
             "MSILHash": "eqmrk1cs/LjuiXQJ+nBw142N/5BaJ8i+unxJ7/AIP0M=",
             "HookCategory": "Entity"
           }
-        }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 21,
+            "ReturnBehavior": 4,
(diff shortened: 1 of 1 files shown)


KrunghCrow:
can already block that look at my StationaryPaper plugin

21mac:
> can already block that look at my StationaryPaper plugin

I did, but I did not see any oxide hook. You use a method called "OnNoteEdit" which is not a hook and therefore does not get called by oxide when a note is changed. You call that function by a console command.
How can this prevent the change by a user?

Substrata1:
Here is an example of how you can do this:
```csharp
object OnServerCommand(ConsoleSystem.Arg arg)
{
    if (arg?.cmd?.FullName == "note.update")
    {
        BasePlayer player = arg.Player();
        string itemid = arg.GetString(0, "");
        string text = arg.GetString(1, "");

        if (text.Contains("bannedword"))
        {
            return true;
        }
    }

    return null;
}
```

MrBlue:
This can be achieved using the OnServerCommand hook as shown above, thanks for your contribution though!

21mac:
Finally tried it.

Works fine, thanks a lot :)

PR #485, opened 2023-12-05
Add CanLockerAcceptItem hook

WheteThunger:
```cs
object CanLockerAcceptItem(Locker locker, Item item, int targetPos)
```
- Called when evaluating whether an item can go into a particular locker slot, at the beginning of the `Locker.ItemFilter` method (assigned to `ItemContainer.canAcceptItem`).
- Different from the `CanAcceptItem` hook, this is called before the `ItemContainer.canAcceptItem` delegate rather than after, and only supports a boolean return value.
- Intended to be used by plugins in conjunction with the `CanWearItem` hook to allow an item to be worn as well as placed in wearable locker slots.

This change also exposes various Locker fields/methods.

Basic Locker functionality was tested before creating this PR, to ensure no regressions.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -19791,6 +19791,30 @@
             "MSILHash": "eqmrk1cs/LjuiXQJ+nBw142N/5BaJ8i+unxJ7/AIP0M=",
             "HookCategory": "Entity"
           }
+        },
+        {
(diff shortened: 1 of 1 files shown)


MrBlue:
Merged, thanks!

PR #486, opened 2023-12-12
Fixing NPE issue of OnPlayerSpawn hook when caller returns non-null

saturna:
Fixing NPE issue of OnPlayerSpawn hook when caller returns non-null

Forum Post Link:
https://umod.org/community/hooks-extended/51527-onplayerspawn-hook-bug-npe?page=1#post-1

Context:
OnPlayerSpawn hook call returns null to the caller if return value is not-null. This will always cause NPE and a player will not be able to spawn onto the server.

In order to allow for SpawnNewPlayer default behavior modifications, BasePlayer needs to be returned instead.

Let me know if you need test evidence.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -9623,7 +9623,7 @@
                 "Operand": 13
               },
               {
-                "OpCode": "ldnull",
+                "OpCode": "ldloc_1",
                 "OpType": "None"
               },
               {


saturna:
Is there a reason why this PR isn't being picked up? It's been several weeks.

MrBlue:
Hey, what happens when you don't initialize the BasePlayer in the hook call. It'll just return a ghost player which I assume will cause issues later on in the joining process no?

If someone returns non-null, is there anything we should be calling before returning the BasePlayer object?

saturna:
If someone doesn't initialize the base player it won't make any difference because that would require to return a non-null value which will cause the exception.

My understanding is that it's up to the caller to initialize BasePlayer however they want - meaning there isn't necessary anything that SHOULD be called. 

If this helps I can dig out my init code from a plugin I wrote using this hook with the patch in place (the proposed DLL patch) that made this hook usable for me. 

Other than that this issue is very easy to reproduce and makes this hook somewhat useless without my proposed fix.

MrBlue:
Merged, thanks!

PR #487, opened 2023-12-16
More performant o.plugins

Tryhard999:
Reduced allocations and overheard from running the O.plugins command, useful for RCON tools and such that might run it often

### DIFF
--- a/src/RustCommands.cs
+++ b/src/RustCommands.cs
@@ -4,6 +4,8 @@ using Oxide.Core.Plugins;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Text;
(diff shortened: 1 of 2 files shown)


MrBlue:
I'm not opposed to changing the way this command works but I feel like we can make it a little bit more readable and clean it up using StringBuilder.AppendFormat for example:

```cs
output.AppendLine();
output.AppendFormat(string.Format("  {0} \"{1}\" ({2}) by {3} ({4}s) - {5}", pluginCount++.ToString("00"), plugin.Title, plugin.Version, plugin.Author, plugin.TotalHookTime.ToString("0.00"), plugin.Filename.Basename()));
```

PR #488, opened 2023-12-17
CuiHelper Json build optimisation.

SkiTee3000:
Json build optimisation.
Added ability to send UI to the list of players.

> According to my tests it reduces serialisation time up to 2-3 times.

### DIFF
--- a/src/RustCui.cs
+++ b/src/RustCui.cs
@@ -6,6 +6,8 @@ using References::Newtonsoft.Json.Converters;
 using References::Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Text;
 using UnityEngine;
 using UnityEngine.UI;
 
@@ -15,11 +17,27 @@ namespace Oxide.Game.Rust.Cui
     {
         public static string ToJson(List<CuiElement> elements, bool format = false)
         {
-            return JsonConvert.SerializeObject(elements, format ? Formatting.Indented : Formatting.None, new JsonSerializerSettings
+            StringBuilder stringBuilder = new StringBuilder();
+            using (StringWriter stringWriter = new StringWriter(stringBuilder))
             {
-                DefaultValueHandling = DefaultValueHandling.Ignore
-            }).Replace("\\n", "\n");
+                using (JsonWriter jsonWriter = new JsonTextWriter(stringWriter))
+                {
+                    jsonWriter.Formatting = Formatting.None;
+
+                    if (elements.Count > 0)
+                    {
+                        jsonWriter.WriteStartArray();
+                        foreach (var element in elements)
+                        {
+                            element.WriteJson(jsonWriter);
+                        }
+                        jsonWriter.WriteEndArray();
+                    }
+                }
+            }
+            return stringBuilder.ToString().Replace("\\n", "\n");
         }
+
 
         public static List<CuiElement> FromJson(string json) => JsonConvert.DeserializeObject<List<CuiElement>>(json);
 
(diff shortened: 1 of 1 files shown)


Vlad-00003:
Correct me if I'm wrong, but this way we loosi 
`DefaultValueHandling = DefaultValueHandling.Ignore`

In this case every field with default value would be send and it would totally mess up `update = true`, because every field that I didn't set manually would be updated to default values

SkiTee3000:
The default values will not be sent. 
I have tested on all UI plugins I use, no problems found.

Vlad-00003:
Oh, my bad
I didn't notice that you're manually checking them in WriteJson

MrBlue:
Hey, this looks interesting but I'm a bit concerned about some of the default value handling. Some such as the AnchorMin/Max you're not sending if the value is set to "0 0" "1 1", however setting those will change the behavior in-game for some reason. Can we test and make sure the variables we're skipping are actually properly being defaulted on the client? The [client code](https://github.com/Facepunch/Rust.Community/blob/master/CommunityEntity.UI.cs) doesn't set the default if the field isn't send in most if not all cases.

SkiTee3000:
> Hey, this looks interesting but I'm a bit concerned about some of the default value handling. Some such as the AnchorMin/Max you're not sending if the value is set to "0 0" "1 1", however setting those will change the behavior in-game for some reason. Can we test and make sure the variables we're skipping are actually properly being defaulted on the client? The [client code](https://github.com/Facepunch/Rust.Community/blob/master/CommunityEntity.UI.cs) doesn't set the default if the field isn't send in most if not all cases.

I have removed this handling of default values, there may indeed be problems when updating the UI.

SkiTee3000:
@MrBlue, could you take a look?

MrBlue:
Hey @SkiTee3000 would you mind updating this again?
Not gonna merge it this wipe but want to try test it before next wipe to make sure there's no issues or breaking changes when we merge it.

SkiTee3000:
> Hey @SkiTee3000 would you mind updating this again? Not gonna merge it this wipe but want to try test it before next wipe to make sure there's no issues or breaking changes when we merge it.

updated

DezLife:
@MrBlue Check It pls

DezLife:
@SkiTee3000 Update this for new cooldown fixes

SkiTee3000:
@DezLife, updated.
@MrBlue, can you check?

DezLife:
> @DezLife, updated. @MrBlue, can you check?

He will add this in an update in the test build. So that it can be tested and then added to the main assembly

SkiTee3000:
I just did a benchmark, compared to the current implementation in Oxide the json build is 3 times faster and there is 10% less memory allocation.
But I can't say that the results of my benchmark are super accurate, it was done on test data, not on any real ui.

DezLife:
> I just did a benchmark, compared to the current implementation in Oxide the json build is 3 times faster and there is 10% less memory allocation. But I can't say that the results of my benchmark are super accurate, it was done on test data, not on any real ui.

This implementation will in any case speed up json serialization and reduce allocations

DezLife:
@SkiTee3000 
Add also the ability to **DestroyUI** for the list of players

DezLife:
@MrBlue  ?

DezLife:
@SkiTee3000  Update for https://github.com/OxideMod/Oxide.Rust/pull/530/

DezLife:
https://github.com/OxideMod/Oxide.Rust/pull/488/commits/4c391e3a2f41637b7f9f68e524055e768d1d6efa

SteamId Must be a string - https://github.com/Facepunch/Rust.Community/blob/master/CommunityEntity.UI.cs#L275

SkiTee3000:
fixed.
I can add my pooling, in which case allocations will be reduced by 3 times, but I need to check if there is an existing pool in oxide so I don't create an extra one.

DezLife:
Now I also noticed that this was moved to RawImage 😂

https://github.com/OxideMod/Oxide.Rust/pull/537

DezLife:
> fixed. I can add my pooling, in which case allocations will be reduced by 3 times, but I need to check if there is an existing pool in oxide so I don't create an extra one.

As far as I know, they did. But I'm not sure that this is suitable for the UI. This may need to be finalized.

But first, it would be good to test what we have now. After this is done you can think about using pools

SkiTee3000:
> > fixed. I can add my pooling, in which case allocations will be reduced by 3 times, but I need to check if there is an existing pool in oxide so I don't create an extra one.
> 
> As far as I know, they did. But I'm not sure that this is suitable for the UI. This may need to be finalized.
> 
> But first, it would be good to test what we have now. After this is done you can think about using pools

I want to apply pooling to a stringbuilder in the ToJson method.

PR #489, opened 2023-12-25
Add OnPatrolHelicopterKill hook

Substrata1:
```csharp
object OnPatrolHelicopterKill(PatrolHelicopter heli, HitInfo hitInfo)
```
Called when the Patrol Helicopter is initially killed, as well as when it finally explodes.
Pre-hook so you can check heli.myAI.isDead to see which it is.


MrBlue:
Merged, thanks!

PR #490, opened 2024-01-07
Change OnResourceDepositCreated ReturnBehavior type

TurEduard:
Now OnResourceDepositCreated can return new ResourceDepositManager.ResourceDeposit in void CreateFromPosition class ResourceDepositManager. This may be necessary to change the properties of the deposit

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -1544,8 +1544,8 @@
         {
           "Type": "Simple",
           "Hook": {
-            "InjectionIndex": 212,
-            "ReturnBehavior": 0,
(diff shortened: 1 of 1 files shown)


MrBlue:
Hey, the `OnResourceDepositCreated` hook is a 'Post' hook which means it's called after something happened and is not intended to change behavior. I'm not entirely sure what you're trying to achieve with the changes you made because they don't seem to change the behavior of the hook at all other than moving the location a little. It might be better to create a 'Pre' hook called `OnResourceDepositCreate` that allows changing the behavior?

PR #493, opened 2024-02-04
Added Hook "CanDropBackpack"

TurEduard:
The hook prevents the backpack from falling out in the Wounded, Crawling, or Dead states.

Upon death, it will still drop if the backpack is not empty, due to the CanAcceptItem check. By changing containerVolumeWhenFilled to zero for ItemModBackpack, you can allow the backpack to move through containers, even if it is not empty.

The hook is also needed to access the backpack when dying, as the backpack will fall out before the OnPlayerDeath hook is triggered.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -17985,6 +17985,30 @@
         {
           "Type": "Simple",
           "Hook": {
+            "InjectionIndex": 9,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
+            "ArgumentString": "l0, this",
+            "HookTypeName": "Simple",
+            "Name": "CanDropBackpack",
+            "HookName": "CanDropBackpack",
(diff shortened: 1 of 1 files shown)


TurEduard:
This is a quick crutch that will prevent a backpack from falling out of a corpse if it has resources. And after that it can only be stuffed into a backpack slot

`private bool CanDropBackpack(Item backpack, PlayerInventory inventory)
{
    ItemModBackpack component = backpack.info.GetComponent<ItemModBackpack>();
    if (component != null)
    {
        component.containerVolumeWhenFilled = 0;
        NextTick(() =>
        {
            component.containerVolumeWhenFilled = 1;
        });
    }

    return false;
}`

MrBlue:
Merged with changes, thanks!
```cs
object OnBackpackDrop(Item backpack, PlayerInventory inventory)
```

PR #496, opened 2024-02-10
Update Command.cs

solorads:
Optimizes it without effecting anything else, It maintains old functions.


MrBlue:
This removes a lot of important logic and seems to be merged improperly, closing but feel free to resubmit a working version!

PR #499, opened 2024-02-13
Add OnNearbyTurretsScan hook

WheteThunger:
```cs
object OnNearbyTurretsScan(AutoTurret turret, List<AutoTurret> nearbyTurrets, bool created)
{
    LogWarning($"OnNearbyTurretsScan works!");
    return null;
}
```

- Called when an auto turret spawns or despawns, in order to determine which turrets are nearby
- Returning non-null will cancel the default Vis.Entities call, allowing for an alternative searching implementation
- The list that will be populated with the Vis.Entities call is passed to the hook, so plugins can simply use a `void` return type and add additional turrets that may not be found by the Vis.Entities call, such as turrets that don't have a collider on the Deployed layer (e.g., for Car Turrets or Drone Turrets)
- Works similar to the `OnSamSiteTargetScan` hook

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -19945,6 +19945,63 @@
             "BaseHookName": "OnPatrolHelicopterTakeDamage",
             "HookCategory": "Entity"
           }
+        },
+        {
+          "Type": "Modify",
+          "Hook": {
+            "InjectionIndex": 2,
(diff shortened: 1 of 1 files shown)


MrBlue:
Merged, thanks!

PR #502, opened 2024-03-09
Add OnItemFilter hook

appendixis:
Added
`object OnItemFilter(StorageContainer container, Item item, int targetSlot)
{
    return null;
}`

Called before check property onlyAcceptCategory in StorageContainer.
Returning true or false overrides default behavior.


MrBlue:
Merged with changes, thanks!
```cs
- OnItemFilter(Item, StorageContainer, int)
```

PR #508, opened 2024-03-24
added OnHackableCrateLaptopAttacked hook

appendixis:
```
object OnHackableCrateLaptopAttacked (HackableLockedCrate crate, HitInfo info)
{
    return null;
}
```

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -20099,6 +20099,30 @@
             "MSILHash": "4UX5r03EqmiQCKgp0C9qRR44ntx2+fpTkb4/sfJaUwA=",
             "HookCategory": "NPC"
           }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 9,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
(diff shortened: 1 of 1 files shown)


Substrata1:
This should work to accomplish the same thing:
```csharp
object OnPlayerAttack(BasePlayer attacker, HitInfo info)
{
    if (info?.HitEntity is HackableLockedCrate && StringPool.Get(info.HitBone) == "laptopcollision")
    {
        Puts("Crate laptop hit!");
    }

    return null;
}
```

appendixis:
> This should work to accomplish the same thing:
> 
> ```cs
> object OnPlayerAttack(BasePlayer attacker, HitInfo info)
> {
>     if (info?.HitEntity is HackableLockedCrate && StringPool.Get(info.HitBone) == "laptopcollision")
>     {
>         Puts("Crate laptop hit!");
>     }
> 
>     return null;
> }
> ```

this called hook on every player attack, but separate hook will call only if attack on laptop

MrBlue:
Merged with changes, thanks!
```cs
- OnCrateLaptopAttack(HackableLockedCrate, HitInfo)
```

PR #510, opened 2024-04-07
Fix OnCollectiblePickedup

DezLife:
The OnCollectiblePickedup call must be made before the item is given to the player


MrBlue:
Merged with changes, thanks!

PR #516, opened 2024-06-17
Adding IOnBaseCombatEntityHurt hook into the Hurt method of the BuildingBlock class

IIIaKa:
Added IOnBaseCombatEntityHurt hook into the Hurt method of the BuildingBlock class, positioned before the PvE mirror damage logic.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -20799,6 +20799,29 @@
             "BaseHookName": "OnCupboardAuthorize [BuildingPrivlidge]",
             "HookCategory": "Structure"
           }
+        },
+        {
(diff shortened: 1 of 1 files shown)


MrBlue:
Hey, 

This would cause the hook to be called twice for BuildingBlocks, plus the BuildingBlock doesn't actually get damaged when the PvE logic gets ran so it wouldn't make sense to have the hook be called there.

Depending on your usecase maybe it might be better to add a hook before the BasePlayer.Hurt method called "OnPvePlayerDamage" or "OnPveProtectionDamage?

PR #520, opened 2024-07-11
Changed some fields/method modifiers of "Bike" to "public"

21mac:
I exposed some fields and methods by changing the modifier to "Public" so they can be used by the plugin system


MrBlue:
Merged with changes, thanks!

PR #523, opened 2024-07-18
Add hook to handle allowing/disallowing debris to spawn for decay entities

cisox:
Hooks into `DecayEntity.SpawnDebris` to allow disabling debris from spawning.


MrBlue:
Merged with changes, thanks!
```cs
- OnDebrisSpawn(DecayEntity, Vector3, Quaternion, bool)
```

PR #525, opened 2024-07-19
Add & modify various AutoTurret hooks

Substrata1:
Existing:
```csharp
object OnTurretModeToggle(AutoTurret turret, BasePlayer player)
```
- Called when a player attempts to toggle the turret mode
- This has been moved, and is now a pre-hook with a player parameter

```csharp
object OnTurretClearList(AutoTurret turret, BasePlayer player)
```
- Called when a player attempts to clear the list of authorized players
- Changed the Argument Behavior from JustParams, which was causing it to not pass the correct parameters, to UseArgumentString

Added:
```csharp
object OnTurretSetID(AutoTurret turret, BasePlayer player, string id)
```
- Called when a player attempts to set the turret's ID

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -4486,7 +4486,7 @@
           "Hook": {
             "InjectionIndex": 12,
             "ReturnBehavior": 1,
-            "ArgumentBehavior": 2,
+            "ArgumentBehavior": 4,
             "ArgumentString": "this, a0.player",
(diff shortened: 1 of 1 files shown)


MrBlue:
Merged with changes, thanks!
```cs
- OnTurretIdentifierSet(AutoTurret turret, BasePlayer player, string id)
```

PR #528, opened 2024-07-22
Using new pluginID field instead of skinID, for modified entites.

IIIaKa:
Hello, I suggest adding a field named pluginID (the name can be changed).

Currently, there are many plugins that use the skinID field for modified entities to mark and/or link them to a plugin.
This is very convenient for other plugins to identify them, but it’s not always possible to specify the plugin's "id" in the skinID. For example, those entities that have skins use this field, and it cannot be used for such purposes.

https://umod.org/community/rust/53529-skinid-pluginid

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -51377,6 +51377,13 @@
           "TypeName": "SprayCanSpray",
           "FieldType": "mscorlib|System.Int32",
           "Flagged": false
+        },
+        {
+          "Name": "pluginID",
+          "AssemblyName": "Assembly-CSharp.dll",
+          "TypeName": "BaseEntity",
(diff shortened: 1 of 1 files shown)


MrBlue:
This feels like too much of a hack, there's too many potential issues such as multiple plugins trying to assign the same variable.
We'll need a better solution to this problem than the one you're providing here unfortunately.

IIIaKa:
> This feels like too much of a hack, there's too many potential issues such as multiple plugins trying to assign the same variable. We'll need a better solution to this problem than the one you're providing here unfortunately.

It is regrettable, as more and more entities are getting their own skins, and the skinID property can no longer be used to 'bind' to the plugin.

PR #534, opened 2024-10-05
Add OnInventoryItemFind and OnInventoryAmmoItemFind hooks

WheteThunger:
```cs
Item OnInventoryItemFind(PlayerInventory inventory, int itemid)
```

- Called when the `PlayerInventory.FindItemByItemID(int itemid)` method is called. For example, when switching ammo types with the currently held weapon, when finding armor to attach to a hot air balloon, when finding a saddle to purchase a horse, and when finding a hood to cover a player's head.
- Returning an `Item` value cancels the default behavior, and causes `PlayerInventory.FindItemByItemID(int itemid)` to return that value
- Works like the [`List<Item> OnInventoryItemsFind(PlayerInventory inventory, int itemid)`](https://github.com/OxideMod/Oxide.Rust/pull/423) hook but for when Rust is searching for only one item instead of a list of items.

```cs
Item OnInventoryAmmoItemFind(PlayerInventory inventory, AmmoTypes ammoType)
```
- Called when the `PlayerInventory.FindAmmo(AmmoTypes ammoType)` method is called. For example, when automatically switching ammo types if the currently selected ammo type is unavailable.
- Returning an `Item` value cancels the default behavior, and causes `PlayerInventory.FindAmmo(AmmoTypes ammoType)` to return that value
- This hook already exists for `Chainsaw` and `FlameThrower` types with the same signature
- This is different from `OnInventoryAmmoFind` in that it returns a single item instead of populating a list

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -18478,7 +18478,7 @@
             "ArgumentBehavior": 4,
             "ArgumentString": "l0.inventory, this.fuelType",
             "HookTypeName": "Simple",
-            "Name": "OnInventoryAmmoItemFind",
+            "Name": "OnInventoryAmmoItemFind [Chainsaw]",
             "HookName": "OnInventoryAmmoItemFind",
             "AssemblyName": "Assembly-CSharp.dll",
             "TypeName": "Chainsaw",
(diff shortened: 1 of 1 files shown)


MrBlue:
Merged, thanks!

PR #539, opened 2024-11-18
HorseDung

IIIaKa:
Adding two hooks to track and optionally cancel Dung creation for **BaseRidableAnimal**.
At the moment, only **RidableHorse** exists.

### 1. The <ins>OnDungProduction</ins> hook is called before the creation of a dung item. Returning a non-null value overrides the default behavior.
```csharp
object OnDungProduction(BaseRidableAnimal animal, ItemDefinition itemDefinition)
{
    Puts($"{animal.GetType()} is about to produce a dung item({itemDefinition.shortname}).");
    return null;
}
```

### 2. The <ins>OnDungProduced</ins> hook is called after the dung item is created and dropped on the ground. No return behavior.
```csharp
void OnDungProduced(BaseRidableAnimal animal, Item item)
{
    Puts($"{animal.GetType()} produced a dung item({item.info.shortname}) in quantity: {item.amount}.");
}
```

#### Code before:
```csharp
private void DoDung()
{
	this.dungProduction -= 1f;
	global::ItemManager.Create(this.Dung, 1, 0uL, true).Drop(base.transform.position + -base.transform.forward + Vector3.up * 1.1f + UnityEngine.Random.insideUnitSphere * 0.1f, -base.transform.forward, default(Quaternion));
}
```

#### Code after:
```csharp
private void DoDung()
{
	this.dungProduction -= 1f;
	if (Interface.CallHook("OnDungProduction", this, this.Dung) != null)
	{
		return;
	}
	global::Item OxideGen_ = global::ItemManager.Create(this.Dung, 1, 0uL, true);
	OxideGen_.Drop(base.transform.position + -base.transform.forward + Vector3.up * 1.1f + UnityEngine.Random.insideUnitSphere * 0.1f, -base.transform.forward, default(Quaternion));
	Interface.CallHook("OnDungProduced", this, OxideGen_);
}
```

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -53646,46 +53646,14 @@
             "HookCategory": "Clan"
           }
         },
-		{
-          "Type": "Modify",
+	{
+          "Type": "Simple",
           "Hook": {
             "InjectionIndex": 6,
-            "RemoveCount": 0,
-            "Instructions": [
-              {
-                "OpCode": "ldstr",
-                "OpType": "String",
-                "Operand": "OnDungProduction"
-              },
-              {
-                "OpCode": "ldarg_0",
-                "OpType": "None"
-              },
-              {
-                "OpCode": "ldarg_0",
-                "OpType": "None"
-              },
-              {
-                "OpCode": "ldfld",
-                "OpType": "Field",
-                "Operand": "Assembly-CSharp|BaseRidableAnimal|Dung"
-              },
-              {
-                "OpCode": "call",
-                "OpType": "Method",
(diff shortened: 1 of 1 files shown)


MrBlue:
Is there a reason you're using Modify hooks instead of simple hooks?
If not can you please update this to use simple hooks instead?

IIIaKa:
> Is there a reason you're using Modify hooks instead of simple hooks? If not can you please update this to use simple hooks instead?
> 
> And please make sure you're adding the hooks in the right place. You're currently adding them to the Rust.Clans.Local dll not Assembly-CSharp

No, there's no particular reason to use Modify. I'll replace them with Simply.

PR #540, opened 2024-12-05
Added OnNpcTarget hook for new NPC types

k1lly0u:


IIIaKa:
This hook is called very, very frequently, unlike the hooks for other NPCs.

IIIaKa:
@MrBlue @k1lly0u 
I think this hook needs to be moved to another method because CanTarget is also called in other places. For example, in the TickTargeting method, within the foreach loop, after CanTarget(item).

PR #546, opened 2025-02-04
Silent item give

IIIaKa:
Currently, when giving items via console commands like **give**, **giveall**, **giveto**, etc., all players receive a notification about it.  
This behavior is appropriate for vanilla servers, where transparency is necessary, but for 99.99% of modded servers, it's a pain in the as*. Most often, such messages are intercepted in the **OnServerMessage** hook to prevent them from being sent, but this approach is inefficient since it requires processing all messages and checking them.

The proposed static property **silentGive**, when set to **true**(default is false), allows suppressing these notifications at the core level, eliminating the need to catch all messages in the **OnServerMessage** hook.

For now, toggling this property is only possible through plugins using **ConVar.Inventory.silentGive = false/true;**.  
It would be great to have this property as an actual ConVar that can be toggled via the console, but as far as I understand, we don’t have access to that.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -21347,6 +21347,252 @@
             "BaseHookName": "OnPlayerVanish",
             "HookCategory": "Player"
           }
+        },
+		{
+          "Type": "Modify",
+          "Hook": {
+            "InjectionIndex": 199,
+            "RemoveCount": 0,
+            "Instructions": [
+              {
+                "OpCode": "ldsfld",
+                "OpType": "Field",
+                "Operand": "Assembly-CSharp|ConVar.Inventory|silentGive"
+              },
+              {
+                "OpCode": "brtrue_s",
+                "OpType": "Instruction",
+                "Operand": 202
+              }
+            ],
+            "HookTypeName": "Modify",
+            "Name": "ItemGiveMsg [give]",
+            "HookName": "ItemGiveMsg",
+            "AssemblyName": "Assembly-CSharp.dll",
+            "TypeName": "ConVar.Inventory",
+            "Flagged": false,
+            "Signature": {
+              "Exposure": 2,
+              "Name": "give",
+              "ReturnType": "System.Void",
+              "Parameters": [
+                "ConsoleSystem/Arg"
+              ]
+            },
+            "MSILHash": "gyMbgBclr3eHbIYGq0QXaPLANKx34WvIc9MXBR/C7RI=",
+            "HookCategory": "_Patches"
+          }
+        },
+        {
+          "Type": "Modify",
+          "Hook": {
+            "InjectionIndex": 117,
+            "RemoveCount": 0,
+            "Instructions": [
+              {
+                "OpCode": "ldsfld",
+                "OpType": "Field",
+                "Operand": "Assembly-CSharp|ConVar.Inventory|silentGive"
+              },
+              {
+                "OpCode": "brtrue_s",
+                "OpType": "Instruction",
+                "Operand": 152
+              }
+            ],
+            "HookTypeName": "Modify",
+            "Name": "ItemGiveMsg [giveall]",
+            "HookName": "ItemGiveMsg",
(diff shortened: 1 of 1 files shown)


MrBlue:
Thanks for submitting this pr, but I don't see the use in the proposed changes. If you want to do this in a more prefrormant way than using the OnServerMessage hook, you should look into using Harmony instead.

IIIaKa:
@MrBlue 
That's exactly the point, to have the ability to disable item give messages without needing a plugin, something like regular ConVar settings.  
Since we can't modify ConsoleSystem.Command[] All, toggling this bool property is currently done through plugins(**ConVar.Inventory.silentGive = false/true;**). However, we could add a command like **o.silentgive true/false** directly in Oxide.

Interestingly, Oxide is mainly used by modded servers, where global chat messages about item give are an issue, players can see admin activity. Everyone uses plugins to block these messages, but using the **OnServerMessage** hook means intercepting and filtering all server messages.

My proposed changes simply add one static bool property and check its value before sending messages.  
I'm sure all modded server owners would support this approach since it reduces the number of operations. And if we also allow toggling it via the console, that would be even better.

**ConVar.Inventory.give(Arg arg):**
```csharp
if (silentGive || basePlayer.IsDeveloper)
{
	if (!flag)
	{
		basePlayer.ChatMessage("you silently gave yourself " + num + " x " + item.info.displayName.english);
	}
	return;
}
Chat.Broadcast(basePlayer.displayName + " gave themselves " + num + " x " + item.info.displayName.english, "SERVER", "#eee", 0uL);
```

**ConVar.Inventory.giveall(Arg arg):**
```csharp
if (!silentGive && item != null)
{
	Chat.Broadcast(text + " gave everyone " + item.amount + " x " + item.info.displayName.english, "SERVER", "#eee", 0uL);
}
```

**ConVar.Inventory.giveto(Arg arg):**
```csharp
if (!silentGive)
{
	Chat.Broadcast(text + " gave " + basePlayer.displayName + " " + num + " x " + item.info.displayName.english, "SERVER", "#eee", 0uL);
}
```

**ConVar.Inventory.giveid(Arg arg):**
```csharp
if (silentGive || basePlayer.IsDeveloper)
{
	basePlayer.ChatMessage("you silently gave yourself " + num + " x " + item.info.displayName.english);
	return;
}
Chat.Broadcast(basePlayer.displayName + " gave themselves " + num + " x " + item.info.displayName.english, "SERVER", "#eee", 0uL);
```

**ConVar.Inventory.givearm(Arg arg):**
```csharp
if (silentGive || basePlayer.IsDeveloper)
{
	basePlayer.ChatMessage("you silently gave yourself " + item.amount + " x " + item.info.displayName.english);
	return;
}
Chat.Broadcast(basePlayer.displayName + " gave themselves " + item.amount + " x " + item.info.displayName.english, "SERVER", "#eee", 0uL);
```

**ConVar.Inventory.copyTo(Arg arg):**
```csharp
if (silentGive || from.IsDeveloper)
{
	from.ChatMessage("you silently copied items to " + toply.displayName);
}
else
{
	Chat.Broadcast(from.displayName + " copied their inventory to " + toply.displayName, "SERVER", "#eee", 0uL);
}
```

**ConVar.Inventory.giveBp(Arg arg):**
```csharp
if (silentGive || basePlayer.IsDeveloper)
{
	basePlayer.ChatMessage("you silently gave yourself 1 x " + item.blueprintTargetDef.shortname + " blueprint");
}
else
{
	Chat.Broadcast(basePlayer.displayName + " gave themselves 1 x " + item.blueprintTargetDef.shortname + " blueprint", "SERVER", "#eee", 0uL);
}
```

IIIaKa:
If anyone is interested, I have created a free plugin that patches all the proposed changes in this PR using Harmony.

https://github.com/IIIaKa/SilentItemsGive
https://umod.org/plugins/silent-items-give
https://codefling.com/plugins/silent-items-give
https://lone.design/product/silent-items-give/

PR #549, opened 2025-02-05
Added OnStartedFreeingContainer(FreeableLootContainer container, BasePlayer player) Hook

dassjosh:
Tested and working


MrBlue:
What is the usecase of this hook exactly?
I would understand adding a hook to the `Release` method to be able to cancel freeing the crate, but this hook isn't cancellable and only gets called when they first start untying the crate?

MrBlue:
Manually merged with changes.

```
- OnFreeableContainerReleaseStarted (FreeableLootContainer, Baseplayer)
```

PR #550, opened 2025-02-25
OnActiveTelephoneUpdated

IIIaKa:
**The <ins>OnActiveTelephoneUpdated</ins> hook is called after the player's active telephone has been changed. The last argument may be *null*.  
No return behavior.**
```csharp
void OnActiveTelephoneUpdated(BasePlayer player, PhoneController phoneController)
{
	Puts($"The active telephone of player '{player.displayName}' has changed to '{(phoneController != null ? phoneController.PhoneNumber : "empty")}'.");
}
```


MrBlue:
Manually merged, thanks!

PR #551, opened 2025-02-25
RidableHorse2

IIIaKa:
With the release of the **February** update, the **RidableHorse** and **BaseRidableAnimal** classes have been deprecated and replaced by **RidableHorse2**. However, all hooks and class fields remained in the old classes, even though the old horses are no longer present on the server.  
This pull request migrates previous modifications and introduces three new hooks.  

### 1. The <ins>OnRidableAnimalTow</ins> hook is called before attempting to tow a horse to siege weapons. Returning a non-null value overrides the default behavior.
```csharp
object OnRidableAnimalTow(RidableHorse2 horse, BasePlayer player)
{
    Puts($"Player '{player.displayName}' is attempting to tow a horse(horse.net.ID.Value) to the siege weapon '{horse.towableEntity.TowEntity.net.ID.Value}'.");
    return null;
}
```

### 2. The <ins>OnRidableAnimalDetach</ins> hook is called before attempting to detach (untow) a horse from a siege weapon. Returning a non-null value overrides the default behavior.
```csharp
object OnRidableAnimalDetach(RidableHorse2 horse, BasePlayer player)
{
    Puts($"Player '{player.displayName}' is attempting to detach a horse({horse.net.ID.Value}) from the siege weapon '{horse.towableEntity.TowEntity.net.ID.Value}'.");
    return null;
}
```

### 3. The <ins>OnSaddleSwap</ins> hook is called before attempting to switch the saddle between single and double. Returning a non-null value overrides the default behavior.
```csharp
object OnSaddleSwap(RidableHorse2 horse, BasePlayer player, Item item)
{
    Puts($"Player '{player.displayName}' is attempting to swap the saddle on the horse({horse.net.ID.Value}) to '{item.info.displayName.english}'.");
    return null;
}
```  

### 4. Fields of the classes that were not modified, as they were not found in RidableHorse2:  
- BaseRidableAnimal::currentVelocity
- BaseRidableAnimal::lastInputTime
- RidableHorse::totalDistance  

### 4. Other changes:  
- Renamed the old hook **OnHorseLead** to **OnRidableAnimalLead**

PR #545, opened 2025-02-04
Non-matching hook names

IIIaKa:
After reviewing the entire list of hooks (for https://github.com/OxideMod/Oxide.Core/pull/103 ), I discovered that some hooks have names that do not correspond correctly to their purpose.  
This pull request is intended to resolve these inconsistencies before they turn into a confusing mess. The sooner this is addressed, the fewer plugins will be affected. **All new names have been verified, do not exist and will not cause conflicts.**

### 1. <ins>Can</ins> hooks that are actually <ins>On</ins> hooks.
**Can** hooks(**bool**) essentially ask plugins whether an action is allowed or not and based on the returned **true** or **false**, the action is either performed or denied(only considering non-null values). The methods that invoke these hooks contain no additional logic beyond checking the return value. This distinguishes them from On(non-object) hooks, which also expect a bool return but serve a different purpose.  
However, for some reason, certain hooks use **Can instead of On**, which creates confusion.
1. CanBuild → **OnBuild**
2. CanCreateWorldProjectile → **OnCreateWorldProjectile**
3. CanAssignBed → **OnSleepingBagAssign**
4. CanLock → **OnLock**(for CodeLock and KeyLock), ModularCarCodeLock remains unchanged
5. CanUnlock → **OnUnlock**(for CodeLock and KeyLock), ModularCarCodeLock remains unchanged
6. CanChangeCode → **OnCodeChange**
7. CanSetBedPublic → **OnSleepingBagPublic**
8. CanMoveItem → **OnItemMove**
9. CanSeeStash → **OnStashReveal**
10. CanHideStash → **OnStashHide**
11. CanCombineDroppedItem → **OnDroppedItemCombine**
12. CanPickupLock → **OnLockPickup**
13. CanUseHelicopter → **OnEntityMount**(CH47HelicopterAIController)
13. CanMountEntity → **OnEntityMount**(BaseMountable)
14. CanDismountEntity → **OnEntityDismount**
15. CanLootEntity → **OnTryLootEntity**(LootableCorpse, ResourceContainer, DroppedItemContainer, StorageContainer, ContainerIOEntity, IndustrialCrafter and WorldItem)
17. CanHackCrate → **OnCrateHack**(existing OnCrateHack renamed to OnCrateHackStart since there is also OnCrateHackEnd)
18. CanRenameBed → **OnSleepingBagRename**
19. CanDeployItem → **OnItemDeploy**
20. CanSamSiteShoot → **OnSamSiteShoot**
[...]

PR #553, opened 2025-03-02
Add missing horse hooks to RidableHorse2

WheteThunger:
- Added hooks from `BaseRidableAnimal` to `RidableHorse2`, including `OnHorseLead`, `CanLootEntity`, and `OnRidableAnimalClaim[ed]`
- Moved existing Horse hooks to the recently added `Animal` category
- Renamed existing Horse hooks to have the `[BaseRidableAnimal]` tag for clarity now that there are two versions

I tested all new hooks.

### DIFF
--- a/resources/Rust.opj
+++ b/resources/Rust.opj
@@ -7717,7 +7717,7 @@
             "ArgumentBehavior": 4,
             "ArgumentString": "this, l0",
             "HookTypeName": "Simple",
-            "Name": "OnHorseLead",
+            "Name": "OnHorseLead [BaseRidableAnimal]",
             "HookName": "OnHorseLead",
             "AssemblyName": "Assembly-CSharp.dll",
             "TypeName": "BaseRidableAnimal",
@@ -7731,7 +7731,32 @@
               ]
             },
             "MSILHash": "BDnxM2qkVLmovqSGo64aramtMcA8Va/YT5U2f7uvBtE=",
-            "HookCategory": "Entity"
+            "HookCategory": "Animal"
+          }
+        },
+        {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 24,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
+            "ArgumentString": "this, l0",
+            "HookTypeName": "Simple",
+            "Name": "OnHorseLead [RidableHorse2]",
+            "HookName": "OnHorseLead",
+            "AssemblyName": "Assembly-CSharp.dll",
+            "TypeName": "RidableHorse2",
+            "Flagged": false,
+            "Signature": {
+              "Exposure": 2,
+              "Name": "SERVER_Lead",
+              "ReturnType": "System.Void",
+              "Parameters": [
+                "BaseEntity/RPCMessage"
+              ]
+            },
+            "MSILHash": "4etJkG6FSFxF9qics6aKI91Dd1LJRG9TqvaA9/uFBxA=",
+            "HookCategory": "Animal"
           }
         },
         {
@@ -8540,6 +8565,31 @@
           }
         },
         {
+          "Type": "Simple",
+          "Hook": {
+            "InjectionIndex": 33,
+            "ReturnBehavior": 1,
+            "ArgumentBehavior": 4,
+            "ArgumentString": "l0, this",
+            "HookTypeName": "Simple",
+            "Name": "CanLootEntity [RidableHorse2]",
+            "HookName": "CanLootEntity",
+            "AssemblyName": "Assembly-CSharp.dll",
+            "TypeName": "RidableHorse2",
+            "Flagged": false,
+            "Signature": {
+              "Exposure": 0,
+              "Name": "SERVER_OpenLoot",
+              "ReturnType": "System.Void",
+              "Parameters": [
+                "BaseEntity/RPCMessage"
+              ]
+            },
+            "MSILHash": "Qk9vfWngOL67eXM5BTQxD64uGdJ1ywhnJKFNtc1TnKw=",
+            "HookCategory": "Player"
+          }
+        },
+        {
           "Type": "Modify",
           "Hook": {
             "InjectionIndex": 117,
@@ -9017,7 +9067,7 @@
               ]
             },
             "MSILHash": "XTfPsD9XIOP0iCGSonevPBtHzykgZULylU5pv8h1tF0=",
-            "HookCategory": "Entity"
+            "HookCategory": "Animal"
           }
         },
         {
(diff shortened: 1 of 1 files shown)


IIIaKa:
@WheteThunger 
Hello, such a PR already exists.
https://github.com/OxideMod/Oxide.Rust/pull/551

WheteThunger:
I see, thanks for pointing it out. That PR appears to focus on adding new hooks, while this PR focuses on solely establishing parity between `BaseRidableAnimal` and `RidableHorse2`. Merging this PR will restore functionality in multiple of my plugins without requiring me to make any changes to them, which is ideal. Glancing at your PR, I see multiple problems. If you can correct those problems, I will close this PR.

- Renaming a hook (`OnHorseLead` -> `OnRidableAnimalLead`) is a backwards-incompatible change
- Removing hooks for `BaseRidableAnimal` is a backwards-incompatible change
- Missing `CanLootEntity` for `RidableHorse2`; I see `OnTryLootEntity` but this is not the name used in `BaseRidableAnimal` or other classes

IIIaKa:
- The `BaseRidableAnimal` and `RidableHorse` classes are deprecated, making it unnecessary to keep changes in them;
 - The change from `OnHorseLead` to `OnRidableAnimalLead` was made to maintain a consistent naming style(`OnRidableAnimalClaim`, `OnRidableAnimalClaimed`, `OnRidableAnimalTow`, `OnRidableAnimalDetach` and etc.).
 - The change from `CanLootEntity` to `OnTryLootEntity` was made because this hook is object, not boolean; `Can` hooks are only for boolean.  

Regarding the changes in naming: `RidableHorse` no longer exists(unless spawned manually), and all plugins will have to update the type from `RidableHorse` to `RidableHorse2`. Therefore, it makes sense to also update the names to more appropriate ones that align with the overall concept.

WheteThunger:
Although the classes are deprecated, I'm not aware of an estimated removal date. If you know it, please share. FP has deprecated things before but has taken their time to remove them. For example, pooling methods, which FP claimed would be removed one month later, but they are still here many months later. I'm not suggesting that anyone develop new plugins or features with the old horses, but existing plugins should be taken into account.

The point of deprecation is to give users time to adjust to the changes before feature removal. Existing pligins should not be broken at random by Oxide maintainers because of an upcoming removal. Assuming that every plugin can react immediately to backwards-incompatible changes suggests a misunderstanding of the plugin ecosystem. Many plugins are not maintained, some were merely authored in a forum post, some were written by Gen AI, etc. I understand that the prefab change for the old horses immediately broke all plugins that relied on knowledge of the prefab, but it was an easy fix (find/replace) even for non-coders, and some plugins did not even need code changes due to the prefab being a config option that the server owner could change. You're right that the old horses no longer spawn naturally in Rust, but some plugins still spawn them, and other plugins need to continue to work with them.

The new horses work differently and some plugins have had to undergo significant changes to restore existing capabilities, so it will take time for the community to upgrade most plugins to support them. It's essential that the community be given sufficient time to make the necessary upgrades. While we are we are beholden to when FP actually removes the old horses, we should not make the problem worse by unnecessarily accelerating that timeline.

Regarding Can vs On hooks, I agree that "new" hooks should follow the convention you described, but I argue this shouldn't be thought of as a new hook. This is backfilling a hook to another class, so it should match the conventions used in other classes.

IIIaKa:
@WheteThunger 
Hello. Well, I think even FP doesn’t know when they'll remove outdated methods and classes :)  
The old horses simply don’t spawn anymore unless manually spawned via a command or plugin. But I don’t think anyone would seriously use them, as the new horses are better, making the old ones obsolete.  
As for existing plugins, they have already "stopped working" since the old horses no longer exist. Any hooks related to them just don’t trigger, even if this hooks still exist. In this case, it might be worth keeping modifications to public methods to prevent compilation errors, but leaving the hooks makes no sense. If this isn't addressed soon, they’ll be forgotten and remain "hanging" indefinitely.  

https://github.com/OxideMod/Oxide.Rust/pull/545
In fact, there are a bunch of hooks with incorrect names. I reviewed all of them while trying to implement conflict avoidance for object hooks. You're right in saying that plugins need time to "adapt", but while we wait for old plugins to "adapt", new plugins are being created daily, increasing the number of plugins using incorrect names.  

Judging by the time that has passed, incorrect hook conflicts have been around for many years and there are no visible signs of progress so far. That's why it's worth at least gradually moving towards solving this problem, starting with the names. For example, new hooks could be added while marking the incorrectly named ones as deprecated, giving developers 2-3 months to transition.  

That’s just my opinion, it may be wrong.

WheteThunger:
Not all existing plugins stopped working. Hooks continue to be called for the old horses. Plugins still referencing the RidableHorse and BaseRidableAnimal will receive the hook calls.

Plugins that explicitly reference the prefab path, such as plugins which spawn horses, were impacted, but as I explained in my previous post, it was a trivial update that even non-coders can do, and I have found that many server owners will take it upon themselves to figure out such changes (often guided by other users in various Discord servers). As I was also saying, some plugins exposed the prefab path as a config option, which allows server owners to correct the problem without a code change.

The Oxide team has renamed hooks before, and what we found is that it's a collective waste of community time. Hundreds of server owners will react to the deprecation notices, resulting in hundreds of independent discussions across various forums (multiple websites, multiple discord servers, many DMs). It simply distracts server owners and developers from spending their time on making truly meaningful changes. The benefit of renaming hooks is very negligible as developers can simply read docs or read the assemblies to learn how to correctly use a hook (which they should do anyway if they are adopting a new hook). Renaming hooks also has a way of forcing old or unmaintained plugins to break, which shouldn't be necessary if those plugins work.

IIIaKa:
Everyone has their own opinion, but I don't think that I, you, or Oxide should be responsible for plugins that are no longer supported. The focus should be on ensuring that existing and future plugins work as intended.  

For example, when working with the TimedPermissions plugin, I was developing a plugin that needed to integrate with it, but I lacked a couple of API methods/hooks. I wanted to request their addition, but then I saw that the plugin had been abandoned for three years. As a result, I simply decided to create my own alternative with more features for both developers and server owners.  
What I want to say is that, no matter how much you worry about plugins abandoned by their authors, their relevance eventually fades. Some plugins simply stop working with such plugins because there is no support from their developers. Meanwhile, new alternatives emerge with active authors who continue to maintain and improve them.  
https://codefling.com/plugins/temporary-permissions

bmgjet:
My 2c, Don't rename hooks. I don't want to have to waste time updating stuff when I don't have to.
The name of the hook means basically nothing to me since I'm looking at the decompiled code and see what hooks I want to use in there. As every one should be doing since all the documentation in the world isn't as straight forward as looking at the code and seeing how a hook works. It could be called GenericHook1 for all I care as long as it works and isn't going to be changed unless it really has to change because some code breaking changes.

IIIaKa:
I see, it's sad that everyone thinks only of themselves...

rostov114:
You can't just rename hooks, you need to mark them as deprecated (src/RustHooks.cs) and give people time to adapt them.
Besides, uMod has quite a few plugins that aren't unmaintained, will you adapt them all to the new hooks?
The idea is good, correct. But there are already few people in the community, and you propose to add a bunch of work to those who remain.

PR #554, opened 2025-03-05
Reopening RidableHorse2

IIIaKa:
From that PR, only the new hooks remain.
https://github.com/OxideMod/Oxide.Rust/pull/551


IIIaKa:
Regarding public methods and properties.  
As you told me, they compile fine, but in the IDE, they cause errors, which creates inconveniences when writing plugins. Is there a way to make them public through the patcher so that the .dlls linked to the IDE also recognize that these properties are accessible?

image
image

bmgjet:
You can set the environment args to dump the publicized files.
Example in bat file for starting server
cd C:\Main
set OXIDE_PublicizerOutput=%~dp0OutTest
RustDedicated.exe -nographics -batchmode

ads102003:
This might be a useful link:
https://docs.oxidemod.com/guides/developers/publicizer

IIIaKa:
@bmgjet @ads102003 
Thank you, I created the '_Patched' folder in the server directory and simply added 'set OXIDE_PublicizerOutput=./_Patched' before 'RustDedicated.exe' in the bat file.

IIIaKa:
As I understand it, no one is going to add data hooks...

PR #574, opened 2026-08-08
Fix permission check for resolved users

BreezeDelegate:
Fixes `oxide.grant user <name> <permission>` checking the resolved player's name instead of their user ID.

The permission check now uses the same resolved `userId` as the grant itself.


ads102003:
Manually merged with a more appropriate patch for a further issue.

PR #575, opened 2026-08-08
Fix DropItem empty slot handling

BreezeDelegate:
Prevents `DropItem(BasePlayer, int)` from accessing item data when an inventory slot is empty.


ads102003:
Thanks for bug report, Did and entire refactor of the method due to other issues found.

